PR #1, opened 2015-06-19
Reading hadoopConfiguration directly from Spark.

Nosfe:
Read hadoopConfiguration from SparkContext instead of creating a new Configuration directly from Hadoop config files.
This allow us to use hadoop parameters inserted or modified in one of Spark's config files. (e.g.: Swift credentials).


yhuai:
@Nosfe Thanks for the PR! I am merging it.

PR #2, opened 2015-07-21
Allow saving benchmark queries results as parquet files

jystephan:
* Add an optional parameter while running the benchmark to specify a directory where the queries results should be saved as parquet files (example use case: make sure the queries results are the same across different benchmark run)
* Remove the hardcoded directory structure baseDir/parquet/ => baseDir/

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/query.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/query.scala
@@ -36,7 +36,7 @@ case class QueryForTest(
     (endTime - startTime).toDouble / 1000000
   }
 
-  def benchmark(description: String = "") = {
+  def benchmark(description: String = "", queryOutputLocation: Option[String]) = {
     try {
       sparkContext.setJobDescription(s"Query: ${query.name}, $description")
       val dataFrame = sqlContext.sql(query.sqlText)
@@ -76,6 +76,8 @@ case class QueryForTest(
           tableIdentifier.last
         }
       }
+
+      queryOutputLocation.foreach(dir => dataFrame.saveAsParquetFile(s"$dir/$name.parquet"))
 
       BenchmarkResult(
         name = query.name,
--- a/src/main/scala/com/databricks/spark/sql/perf/query.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/query.scala
@@ -90,7 +90,7 @@ case class QueryForTest(
           tableIdentifier.last
         }
       }
-      
+
       BenchmarkResult(
         name = query.name,
         joinTypes = joinTypes,
--- a/src/main/scala/com/databricks/spark/sql/perf/query.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/query.scala
@@ -78,9 +78,7 @@ case class QueryForTest(
         query.executionMode match {
           case CollectResults => dataFrame.rdd.collect()
           case ForeachResults => dataFrame.rdd.foreach { row => Unit }
-          case WriteParquet(location) => {
-            dataFrame.rdd.collect()
-            dataFrame.saveAsParquetFile(s"$location/$name.parquet")
+          case WriteParquet(location) => dataFrame.saveAsParquetFile(s"$location/$name.parquet")
           }
         }
       }
--- a/src/main/scala/com/databricks/spark/sql/perf/query.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/query.scala
@@ -19,7 +19,21 @@ package com.databricks.spark.sql.perf
 import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.catalyst.analysis.UnresolvedRelation
 
-case class Query(name: String, sqlText: String, description: String, collectResults: Boolean)
+/**
+ * The execution mode of a benchmark:
+ * - CollectResults: Benchmark run by collecting queries results
+ *   (e.g. rdd.collect())
+ * - ForeachResults: Benchmark run by iterating through the queries results rows
+ *   (e.g. rdd.foreach(row => Unit))
+ * - WriteParquet(location): Benchmark run by saving the output of each query as a
+ *   parquet file at the specified location
+ */
+abstract class ExecutionMode
+case object CollectResults extends ExecutionMode
+case object ForeachResults extends ExecutionMode
+case class WriteParquet(location: String) extends ExecutionMode
+
+case class Query(name: String, sqlText: String, description: String, executionMode: ExecutionMode)
 
 case class QueryForTest(
     query: Query,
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/spark/sql/perf/query.scala
```diff
@@ -36,7 +36,7 @@ case class QueryForTest(
     (endTime - startTime).toDouble / 1000000
   }
 
-  def benchmark(description: String = "") = {
+  def benchmark(description: String = "", queryOutputLocation: Option[String]) = {
```
marmbrus: Could we instead do something like the following?

```scala
abstract class ExecutionMode
case object CollectResults extends ExecutionMode
case object CountResults extends ExecutionMode
case object WriteParquet(location: String) extends ExecutionMode
```

This can replace the `collectResults` boolean and the query itself can tell the harness how it should be run.

jystephan: Thanks -- working on it!

src/main/scala/com/databricks/spark/sql/perf/query.scala
```diff
@@ -60,10 +74,10 @@ case class QueryForTest(
...
 
       // The executionTime for the entire query includes the time of type conversion from catalyst to scala.
-      val executionTime = if (query.collectResults) {
-        benchmarkMs { dataFrame.rdd.collect() }
-      } else {
-        benchmarkMs { dataFrame.rdd.foreach {row => Unit } }
+      val executionTime = query.executionMode match {
+        case CollectResults => benchmarkMs { dataFrame.rdd.collect() }
+        case ForeachResults => benchmarkMs { dataFrame.rdd.foreach { row => Unit } }
+        case WriteParquet(location) => benchmarkMs { dataFrame.saveAsParquetFile(s"$location/$name.parquet") }
```
aarondav: 100ch

aarondav: maybe wrap the whole thing by benchmarkMs.

src/main/scala/com/databricks/spark/sql/perf/query.scala
```diff
@@ -19,7 +19,21 @@ package com.databricks.spark.sql.perf
...
+ * The execution mode of a benchmark:
+ * - CollectResults: Benchmark run by collecting queries results
+ *   (e.g. rdd.collect())
+ * - ForeachResults: Benchmark run by iterating through the queries results rows
+ *   (e.g. rdd.foreach(row => Unit))
+ * - WriteParquet(location): Benchmark run by saving the output of each query as a
+ *   parquet file at the specified location
+ */
+abstract class ExecutionMode
+case object CollectResults extends ExecutionMode
```
aarondav: Let's put the actual objects inside of an object like 
```scala
trait ExecutionMode
object ExecutionMode {
  case object CollectResults extends Executionmode
}
```

so we can refer to it by its fully qualified name ExecutionMode.CollectResults from outside of the class, I think this makes the code more readable from the API perspeective.

src/main/scala/com/databricks/spark/sql/perf/tpcds/queries/ImpalaKitQueries.scala
```diff
@@ -1463,7 +1463,7 @@ object ImpalaKitQueries {
         |from store_sales
       """.stripMargin)
   ).map {
-    case (name, sqlText) => Query(name, sqlText, description = "original query", collectResults = true)
+    case (name, sqlText) => Query(name, sqlText, description = "original query", executionMode = CollectResults)
```
aarondav: 100ch

src/main/scala/com/databricks/spark/sql/perf/tpcds/queries/SimpleQueries.scala
```diff
@@ -137,6 +137,6 @@ object SimpleQueries {
               |-- end query 1 in stream 0 using template query7.tpl
             """.stripMargin)
    ).map {
-     case (name, sqlText) => Query(name = name, sqlText = sqlText, description = "", collectResults = false)
+     case (name, sqlText) => Query(name = name, sqlText = sqlText, description = "", executionMode = ForeachResults)
```
aarondav: 100ch

src/main/scala/com/databricks/spark/sql/perf/table.scala
```diff
@@ -47,7 +35,7 @@ abstract class TableForTest(
 
   val name = table.name
 
-  val outputDir = s"$baseDir/parquet/${name}"
+  val outputDir = s"$baseDir/${name}"
```
aarondav: Why this change?

src/main/scala/com/databricks/spark/sql/perf/query.scala
jystephan: Thanks, working on it!

src/main/scala/com/databricks/spark/sql/perf/table.scala
jystephan: The hardcoded structure with parquet is not respected by all datasets (in particular for the kms dataset that we have internally) so it was handy to be able to specify directly the path we wanted at TPCDS creation.

aarondav: Gotcha, just gotta make sure this doesn't break any existing infra :)

jystephan: Good point! @yhuai @marmbrus ? 
The alternative being that we could make sure the kms-encrypted dataset has the baseDir/parquet/ structure. So that this change is not needed.

marmbrus: This change is fine.  I'm going to redo all of the data generation at some point anyway now that partitioning is a first class citizen.

src/main/scala/com/databricks/spark/sql/perf/query.scala
```diff
@@ -59,11 +72,17 @@ case class QueryForTest(
...
-        benchmarkMs { dataFrame.rdd.foreach {row => Unit } }
+      // The executionTime for the entire query includes the time of type conversion
+      // from catalyst to scala.
+      val executionTime = benchmarkMs {
+        query.executionMode match {
+          case CollectResults => dataFrame.rdd.collect()
+          case ForeachResults => dataFrame.rdd.foreach { row => Unit }
+          case WriteParquet(location) => {
+            dataFrame.rdd.collect()
+            dataFrame.saveAsParquetFile(s"$location/$name.parquet")
```
yhuai: If we do `collect` first and then do `saveAsParquetFile`, we will execute the query twice. I think we can remove `dataFrame.rdd.collect()`. Also, I guess it is alternative option of `ForeachResults` to benchmark queries having many output rows. So, I think we'd better not to call collect, which will retrieve all results to the driver.

yhuai:
Just one comment. Otherwise, it looks good!

yhuai:
Thanks! I am merging it.

PR #4, opened 2015-07-24
Job does not submit in TPCDS partitioned table data generate

romansew:
Job does not submit in TPCDS partitioned table data generate
The table "store_sales" data result path is
/tmp/spark_perf/scaleFactor=1/useDecimal=true/store_sales/ss_sold_date_sk=2452822/_temporary/0/task_201507241546_0000_r_000000/part-r-00000.parquet

in spark1.3.1 the "_temporary" path is ok. but in spark1.4+ "_temporary" path is filtered.so the exception when query:
java.lang.AssertionError: assertion failed: No schema defined, and no Parquet data file or summary file found under /tmp/spark_perf/scaleFactor=1/useDecimal=true/store_sales.

while the job commit the partitioned data path is
/tmp/spark_perf/scaleFactor=1/useDecimal=true/store_sales/ss_sold_date_sk=2452822/part-r-00000.parquet

the data read in spark1.3+ is ok.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
@@ -108,7 +108,7 @@ case class TPCDSTableForTest(
         val job = new Job(sqlContext.sparkContext.hadoopConfiguration)
 
         val writeSupport =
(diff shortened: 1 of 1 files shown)


marmbrus:
Thanks for the patch, but we have since switched this code to just use the native partition writing that was added in Spark 1.4.

PR #6, opened 2015-08-11
Remove deprecated parquet writing code and add some micro benchmarks

marmbrus:
This PR contains several improvements and also removes the very hacky parquet partition writing code (as it no longer compiles with Spark 1.5 and this functionality is implemented correctly in this version).  I also did some refactoring to make it easier to write traits that contain specific micro benchmarks.  See https://github.com/databricks/spark-sql-perf/compare/databricks:master...marmbrus:refactor?expand=1#diff-c5a813caa37f152168cb6efd9838908eR5 as an example


yhuai:
Merging!

PR #7, opened 2015-08-13
Fixes to breakdown calculation and table creation.

marmbrus:


src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -413,27 +419,47 @@ abstract class Benchmark(@transient protected val sqlContext: SQLContext)
...
+          val indexMap = physicalOperators.map { case (index, op) => (op, index) }.toMap
+          val timeMap = new mutable.HashMap[Int, Double]
+
+          physicalOperators.reverse.map {
             case (index, node) =>
+              messages += s"Breakdown: ${node.simpleString}"
+              val newNode = buildDataFrame.queryExecution.executedPlan(index)
               val executionTime = benchmarkMs {
-                node.execute().map(_.copy()).foreach(row => Unit)
+                node.execute().foreach((row: Any) => Unit)
```
yhuai: `newNode`?

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -413,27 +419,47 @@ abstract class Benchmark(@transient protected val sqlContext: SQLContext)
...
               }
-              BreakdownResult(node.nodeName, node.simpleString, index, executionTime)
+              timeMap += ((index, executionTime))
+
+              val childIndexes = node.children.map(indexMap)
```
yhuai: `newNode`?

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -413,27 +419,47 @@ abstract class Benchmark(@transient protected val sqlContext: SQLContext)
...
+              val childTime = childIndexes.map(timeMap).sum
+
+              messages += s"Breakdown time: $executionTime (+${executionTime - childTime})"
+
+              BreakdownResult(
+                node.nodeName,
+                node.simpleString.replaceAll("#\\d+", ""),
```
yhuai: `newNode`?

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
marmbrus: good catch!

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
marmbrus: Here I think we actually want to use the original in case attribute ids or anything change amongst trees.

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
marmbrus: same

yhuai:
Merging!

PR #8, opened 2015-08-14
Add a new ExecutionMode to calculate the sum of hash values of result rows.

yhuai:
We save this value in the result json file, which can be used to check query results.


src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -66,6 +67,7 @@ case class BenchmarkResult(
     optimizationTime: Option[Double] = None,
     planningTime: Option[Double] = None,
     executionTime: Option[Double] = None,
+    hashSum: Option[Long] = None,
```
marmbrus: maybe just make this `result`? I was thinking other execution modes might do things like return num rows. (though then maybe we should also include the execution mode in the result?)

PR #17, opened 2015-09-10
Add more tests for join performance

marmbrus:


marmbrus:
Discussed offline w/ @rxin, merging.

PR #20, opened 2015-09-10
Add a plugin for making releases to Databricks

marmbrus:
Also releases 0.1.2.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,7 +61,7 @@ releaseProcess := Seq[ReleaseStep](
   commitReleaseVersion,
   tagRelease,
   setupDbcRelease,
-  releaseStepTask(dbcDeploy),
+  releaseStepTask(dbcUpload),
   setNextVersion,
   commitNextVersion,
   pushChanges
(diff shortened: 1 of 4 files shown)


yhuai:
Do not use merge script for this one.

build.sbt
```diff
@@ -23,4 +21,48 @@ initialCommands in console :=
...
+// Add publishing to spark packages as another step.
+releaseProcess := Seq[ReleaseStep](
+  checkSnapshotDependencies,
+  inquireVersions,
+  runTest,
+  setReleaseVersion,
+  commitReleaseVersion,
+  tagRelease,
+  setupDbcRelease,
+  releaseStepTask(dbcDeploy),
```
yhuai: Maybe use `dbcUpload` instead of `dbcDeploy` because `dbcDeploy` will try to attach to a cluster. If the cluster does not exist, the release process will fail.

yhuai: Or, we can ask if the person making the release if we should attach to a cluster?

yhuai:
I tried it out and it works great (I did not push changes to remove though).

PR #22, opened 2015-09-16
Add 2 queries

bit1129:


src/main/scala/com/databricks/spark/sql/perf/tpcds/SimpleQueries.scala
```diff
@@ -138,7 +138,38 @@ trait SimpleQueries extends Benchmark {
               |  i_item_id
               |limit 100
               |-- end query 1 in stream 0 using template query7.tpl
-            """.stripMargin)
+            """.stripMargin),
+            
+      ("q7-bigtable-selfjoin-1",   """
```
yhuai: Since it is not very related to q7, I'd change the name to "store_sales-selfjoin-1".

src/main/scala/com/databricks/spark/sql/perf/tpcds/SimpleQueries.scala
```diff
@@ -138,7 +138,38 @@ trait SimpleQueries extends Benchmark {
...
+                                   |-- 1.5 is slower than 1.4.1
```
yhuai: Instead of mentioning that 1.5 is slower than 1.4.1, it will be helpful to mention that this query will generate much more rows than input rows because the join condition will yield many matches.

yhuai:
two minor comments. I will make the changes and merge it. Thanks!

PR #28, opened 2015-11-12
Child tables

nongli:
In some of the tpcds tables, there is a parent/child relationship. This means that
the child table cannot be generated independently and is output (optionally) as part
of generating the parent table. This benchmark requires the output being written to
stdout. When the tpcds generator runs with stdout as the output, it cannot support
child tables.

The dsdgen tool has been updated to support this and can optionally output the data
with the table prefix. For example, in this mode the output is:

store_sales|<row>
store_sales|<row>
store_returns|<row>
store_sales|<row>

The rows are interleaved but can be filtered for. This patch adds the filtering.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
@@ -16,15 +16,12 @@
 
 package com.databricks.spark.sql.perf.tpcds
 
-import java.io.File
-
-import scala.sys.process._
-
 import org.apache.spark.Logging
-import org.apache.spark.sql.{SaveMode, SQLContext}
-import org.apache.spark.sql.Row
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types._
+import org.apache.spark.sql.{DataFrame, Row, SQLContext, SaveMode}
+
+import scala.sys.process._
 
 class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int) extends Serializable with Logging {
   import sqlContext.implicits._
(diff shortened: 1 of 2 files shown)


nongli:
Does the commit message show up in github?

    In some of the tpcds tables, there is a parent/child relationship. This means that
    the child table cannot be generated independently and is output (optionally) as part
    of generating the parent table. This benchmark requires the output being written to
    stdout. When the tpcds generator runs with stdout as the output, it cannot support
    child tables.

    The dsdgen tool has been updated to support this and can optionally output the data
    with the table prefix. For example, in this mode the output is:

    store_sales|<row>
    store_sales|<row>
    store_returns|<row>
    store_sales|<row>

    The rows are interleaved but can be filtered for. This patch adds the filtering.

marmbrus:
If you put it in the description of the PR, the merge script will make it the commit message when we merge.

marmbrus:
Oh, and you can click `...` to see the full message for any given commit before we merge.

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
```diff
@@ -32,15 +29,30 @@ class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int) extend
   def sparkContext = sqlContext.sparkContext
   val dsdgen = s"$dsdgenDir/dsdgen"
 
+  /**
+    * For some tables, it is not generated on its own but instead when the parent table is
+    * generated. Returns the parent table for these kind of tables.
+    */
+  def getParentTable(tbl: String): Option[String] = {
```
marmbrus: This could probably also be:

```scala
val parentTables = Map("store_returns" -> "store_sales")
parentTables.get(...) // Option[String]
```

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
```diff
@@ -53,43 +65,71 @@ class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int) extend
...
           val commands = Seq(
             "bash", "-c",
-            s"cd $localToolsDir && ./dsdgen -table $name -filter Y -scale $scaleFactor -RNGSEED 100 $parallel")
+            s"cd $localToolsDir && ./dsdgen -table $table -filter Y -scale $scaleFactor -RNGSEED 100 $parallel $prefix")
           println(commands)
           commands.lines
         }
       }
 
+      if (parentTable.isDefined) {
```
nongli: Davies is working on a better way to do this. I'd hold off looking at this any more. I think we can just make all the changes in dsdgen and leave this as is.

PR #29, opened 2015-11-13
Update to Spark 1.6

marmbrus:
Some internal interfaces changed, so we need to bump the Spark version to run tests on Spark 1.6.


marmbrus:
Note that this is going to break travis until we publish a preview release.

PR #33, opened 2015-11-17
Increase integration surface area with Spark perf

andrewor14:
The changes in this PR are centered around making `Benchmark#runExperiment` accept things other than `Query`s. In particular, in spark-perf we don't always have a DataFrame or an RDD to work with and may want to run arbitrary code (e.g. ALS.train). This PR makes it possible to use the same code in `Benchmark` to do this.

I tested this on dogfood and it works well there.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
@@ -92,20 +92,20 @@ abstract class Benchmark(
   }
 
   /**
-   * Starts an experiment run with a given set of queries.
-   * @param queriesToRun a list of queries to be executed.
-   * @param includeBreakdown If it is true, breakdown results of a query will be recorded.
+   * Starts an experiment run with a given set of executions to run.
+   * @param executionsToRun a list of executions to run.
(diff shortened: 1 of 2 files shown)


andrewor14:
@marmbrus please have a look

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -92,20 +93,20 @@ abstract class Benchmark(
...
    * @param variations [[Variation]]s used in this run.  The cross product of all variations will be
-   *                   run for each query * iteration.
+   *                   run for each execution * iteration.
    * @param tags Tags of this run.
    * @return It returns a ExperimentStatus object that can be used to
    *         track the progress of this experiment run.
    */
   def runExperiment(
-      queriesToRun: Seq[Query],
+      executionsToRun: Seq[Benchmarkable],
```
marmbrus: BTW, in the future we should avoid changing argument names, even if they aren't totally correct.  This broke a bunch of notebooks that were using named arguments.

andrewor14: ok

PR #38, opened 2015-11-24
Fixes for Q34 and Q73 to return results deterministically.

nongli:


gatorsmile:
@nongli @marmbrus 
Have you tried all the other queries? Have you tested Spark 1.6.0 on all the TPCDS queries? Thank you!

gatorsmile:
Never mind. I just got the voting email from @marmbrus . It sounds like all the cases have already passed in 1.6.0. Thank you!

PR #40, opened 2015-12-02
Add Streaming perf support

zsxwing:
Add `StreamingPerfExecution` to support running Streaming perf jobs and collecting more than one results.


src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -74,6 +75,7 @@ case class BenchmarkResult(
     planningTime: Option[Double] = None,
     executionTime: Option[Double] = None,
     result: Option[Long] = None,
+    doubleResults: Map[String, Double] = Map.empty,
```
marmbrus: Can you describe more what you are doing here?  Once we add things to this schema we are stuck with them forever.

zsxwing: Streaming perf jobs need to collect multiple results and may add more later. And we will encounter 22 fields issue if just adding all results to different fields.

zsxwing:
CC @tdas

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -76,7 +77,21 @@ case class BenchmarkResult(
...
+ * @param schedulingDelay The scheduling delay results of a Streaming test. The key is something like, avg, min, max,
+ *                        25%, 50%...
+ * @param processingDelay The processing delay results of a Streaming test. The key is something like, avg, min, max,
+ *                        25%, 50%...
+ * @param totalDelay The total delay results of a Streaming test. The key is something like, avg, min, max, 25%, 50%
+ */
+case class StreamingBenchmarkResult(
+    schedulingDelay: Map[String, Double],
+    processingDelay: Map[String, Double],
+    totalDelay: Map[String, Double])
```
andrewor14: total delay is just scheduling + processing right? Do we need it?

andrewor14:
LGTM, can you resolve the conflicts? @marmbrus any other comments?

zsxwing:
Closing this one. After discussing with @tdas, we decide to move streaming-perf tests to this project and create better case classes for Streaming results.

PR #41, opened 2015-12-03
Make ExecutionMode.HashResults handle null value

yhuai:
In Spark 1.6, if a value is null, `getLong` will throw an exception. Before 1.6, it will return 0. With this PR, we will check if the result is null. If it is null, null will be returned instead of 0.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
@@ -597,7 +597,7 @@ abstract class Benchmark(
         // to scala.
         // The executionTime for the entire query includes the time of type conversion
         // from catalyst to scala.
-        var result: Option[java.lang.Long] = None
+        var result: Option[Long] = None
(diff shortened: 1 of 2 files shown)


yhuai:
@marmbrus or @nongli  take a look?

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -597,7 +597,7 @@ abstract class Benchmark(
         // to scala.
         // The executionTime for the entire query includes the time of type conversion
         // from catalyst to scala.
-        var result: Option[Long] = None
+        var result: Option[java.lang.Long] = None
```
marmbrus: Its a little odd for this be both an `Option` and a boxed `java.lang.Long`.  Can we pick one or the other?  Having a `Some(null)` is really confusing.

yhuai: oh, i see. Let's just use Option.

PR #47, opened 2015-12-23
Main Class for running Benchmarks from the command line

marmbrus:
This PR adds the ability to run performance test locally as a stand alone program that reports the results to the console:

```
$ bin/run --help
spark-sql-perf 0.2.0
Usage: spark-sql-perf [options]

  -b <value> | --benchmark <value>
        the name of the benchmark to run
  -f <value> | --filter <value>
        a filter on the name of the queries to run
  -i <value> | --iterations <value>
        the number of iterations to run
  --help
        prints this usage text


$ bin/run --benchmark DatasetPerformance
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@ This is a performance testing framework for [Spark SQL](https://spark.apache.org
 # Quick Start
 
 ```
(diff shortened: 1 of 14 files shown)


yhuai:
LGTM. Tried it locally with `DatasetPerformance`. Tests were good and I can see the result dir `performance`.

PR #53, opened 2016-02-23
Extract Query class from Benchmark into its own top-level class and make SparkContext field transient

JoshRosen:
This patch extracts `Query` into its own top-level class and makes its `sparkContext` field transient in order to fix `NotSerializableException`s.


JoshRosen:
Whoops, looks like this didn't entirely fix the serialization issue, so please don't merge yet.

JoshRosen:
Alright, updated and this new commit works.

marmbrus:
LGTM

JoshRosen:
Sweet, I'll merge this then.

PR #69, opened 2016-06-20
Initial commit for adding MLlib reporting in spark-sql-perf

thunterdb:
This PR adds basic MLlib infrastructure to run some benchmarks against ML pipelines.

There are 2 ways to describe and run ML pipelines:
 - programatically, in scala (see MLBenchmarks.scala)
 - using a simple YAML file (see mllib-small.yaml for an example)
The YAML approach is preferred because it generates programmatically the cartesian product of all the experiments to run and validates the types of the objects in the yaml file.

In both cases, all the ML experiments are standard benchmarks.

This PR also moves some code in `Benchmark.scala` : the current code generates path-dependent structural signatures and confuses intellij.

It does not include tests, but some small benchmarks can be run locally against a spark 2 installation:

```
$SPARK_HOME/bin/spark-shell --jars $PWD/target/scala-2.10/spark-sql-perf-assembly-0.4.9-SNAPSHOT.jar
```
and then:

```scala
com.databricks.spark.sql.perf.mllib.MLLib.run(yamlFile="src/main/scala/configs/mllib-small.yaml")
```


thunterdb:
cc @jkbradley

davies:
cc @yhuai

jkbradley:
I'll review the ML parts, but it would be great to get @yhuai 's feedback about the changes to:
* build.sbt
* shared components
* (ML elements if you have time)

src/main/scala/com/databricks/spark/sql/perf/mllib/MLAlgorithmTests.scala
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.spark.SparkContext
+import org.apache.spark.ml.classification.{LogisticRegression, LogisticRegressionModel}
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{DataFrame, Row, SQLContext, Dataset}
+
+import scala.collection.mutable.ArrayBuffer
+
+
+/** Parent class for tests which run on a large dataset. */
+abstract class RegressionAndClassificationTests[M](
```
jkbradley: To confirm: This entire file is outdated and should be removed, right?

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -122,238 +103,14 @@ abstract class Benchmark(
...
-                      query.newDataFrame().queryExecution.executedPlan.toString()
-                    } catch {
-                      case e: Exception =>
-                        s"failed to parse: $e"
-                    }
-                  case _ => ""
-                }
-                startTime = System.currentTimeMillis()
+      timeout: Long = 0L,
+      resultLocation: String = resultsLocation,
```
jkbradley: Shouldn't this be set via the sqlContext conf?

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
```diff
@@ -35,10 +39,16 @@ trait Benchmarkable {
...
       description: String = "",
       messages: ArrayBuffer[String],
-      timeout: Long): BenchmarkResult = {
+      timeout: Long,
+      forkThread: Boolean = true): BenchmarkResult = {
+    logger.info(s"$this: benchmark")
     sparkContext.setJobDescription(s"Execution: $name, $description")
     beforeBenchmark()
-    val result = runBenchmark(includeBreakdown, description, messages, timeout)
+    val result = if (forkThread) {
```
jkbradley: Is this if-then-else switched?  Isn't it forking when it should not?

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
```diff
@@ -35,10 +39,16 @@ trait Benchmarkable {
...
+      doBenchmark(includeBreakdown, description, messages)
+    } else {
+      runBenchmarkForked(includeBreakdown, description, messages, timeout)
```
jkbradley: TODO: Why do we need to avoid forking?

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
```diff
@@ -93,4 +111,11 @@ trait Benchmarkable {
     val endTime = System.nanoTime()
     (endTime - startTime).toDouble / 1000000
   }
-}
\ No newline at end of file
+
+  protected def measureTime[A](f: => A): (Duration, A) = {
```
jkbradley: Make time scale obvious in method name

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -0,0 +1,94 @@
...
+ *  - generate a model against the training data
+ *  - score the model against the training data
+ *  - score the model against the test data
+ *
+ * You should not assume that your implementation can carry state around. If some state is needed,
+ * consider adding it to the context.
+ *
+ * It is assumed that the implementation is going to be an object.
+ */
+trait BenchmarkAlgorithm extends Logging {
```
jkbradley: TODO: REVIEW

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  // This is internal to the implementation of the benchmark.
+  type Model <: Transformer
```
jkbradley: Remove, and just use ```Model[_]``` or ```Transformer```.  The latter might be good if we want to test Transformers later on.

src/main/scala/com/databricks/spark/sql/perf/mllib/ClassificationContext.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.spark.sql.SQLContext
+
+
+/**
+ * All the information required to run a test.
+ * @param commonParams
+ * @param extraParams
+ * @param sqlContext
+ */
+// TODO(tjh) rename, this is not only about classification
```
jkbradley: noting this TODO.  Also, isn't this almost the same as

src/main/scala/com/databricks/spark/sql/perf/mllib/ClassificationContext.scala
```diff
@@ -0,0 +1,41 @@
...
+      new java.util.Random().nextInt()
+    }
+  }
+
+  /**
+   * A fixed seed for this class. This function will always return the same value.
+   *
+   * @return
+   */
+  def seed(): Int = internalSeed
```
jkbradley: Use Long

src/main/scala/com/databricks/spark/sql/perf/mllib/MLBenchmarks.scala
```diff
@@ -0,0 +1,39 @@
+package com.databricks.spark.sql.perf.mllib
+
+import com.databricks.spark.sql.perf.mllib.classification.LogisticRegression
+import org.apache.spark.SparkContext
+import org.apache.spark.sql.SQLContext
+
+import com.databricks.spark.sql.perf.{ExtraMLTestParameters, MLTestParameters}
+import OptionImplicits._
+
+case class MLBenchmark(
```
jkbradley: I'd prefer to rename this since "Benchmark" has a special meaning already in spark-sql-perf.  Maybe we can call this concept a "Test" instead of a "Benchmark."

src/main/scala/com/databricks/spark/sql/perf/mllib/MLClassificationBenchmarkable.scala
```diff
@@ -0,0 +1,110 @@
+package com.databricks.spark.sql.perf.mllib
+
+import com.databricks.spark.sql.perf._
+import com.typesafe.scalalogging.slf4j.Logging
+import org.apache.spark.sql._
+
+import scala.collection.mutable.ArrayBuffer
+
+// TODO(tjh) rename, this is not only about classification
```
jkbradley: Noting this TODO

src/main/scala/com/databricks/spark/sql/perf/mllib/MLClassificationBenchmarkable.scala
```diff
@@ -0,0 +1,110 @@
...
+      trainingData.cache()
+      trainingData.count()
+    } catch {
+      case e: Throwable =>
+        println(s"$this error in beforeBenchmark: ${e.getStackTraceString}")
+        throw e
+    }
+  }
+
+  override def doBenchmark(
```
jkbradley: Is there a need to make this public and not protected?  Same for other methods/vals.

src/main/scala/com/databricks/spark/sql/perf/mllib/MLLib.scala
```diff
@@ -0,0 +1,62 @@
...
+
+
+class MLLib(@transient sqlContext: SQLContext)
+  extends Benchmark(sqlContext) with Serializable {
+
+  def this() = this(SQLContext.getOrCreate(SparkContext.getOrCreate()))
+}
+
+object MLLib extends Logging {
+  def runDefault(runConfig: RunConfig): MLLib = {
```
jkbradley: Document

src/main/scala/com/databricks/spark/sql/perf/mllib/OptionImplicits.scala
```diff
@@ -0,0 +1,39 @@
+package com.databricks.spark.sql.perf.mllib
+
+import scala.language.implicitConversions
+
+object OptionImplicits {
```
jkbradley: Document what this is used for.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.spark.ml.evaluation.MulticlassClassificationEvaluator
+import org.apache.spark.ml.linalg.Vectors
+import org.apache.spark.sql.{DataFrame}
+
+object LogisticRegression extends BenchmarkAlgorithm
+  with TestFromTraining with TrainingSetFromTransformer with ScoringWithEvaluator {
+
+  override type Model = LogisticRegressionModel
+
+  def evaluator(ctx: ClassificationContext) = new MulticlassClassificationEvaluator()
```
jkbradley: I'd use this same evaluator for all classification tests.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def initialData(ctx: ClassificationContext) = {
```
jkbradley: Can you make some things private?  It helps with review.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
```diff
@@ -0,0 +1,49 @@
...
+    import ctx.commonParams._
+    DataGenerator.generateFeatures(
+      ctx.sqlContext,
+      numExamples,
+      ctx.seed(),
+      numPartitions,
+      numFeatures)
+  }
+
+  def initialModel(ctx: ClassificationContext): Model = {
```
jkbradley: trueModel

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  def train(ctx: ClassificationContext,
+            trainingSet: DataFrame): Model = {
+    logger.info(s"$this: train: trainingSet=${trainingSet.schema}")
+    import ctx.extraParams._
+    val lr = new LogisticRegression()
+      .setTol(tol)
+      .setRegParam(regParam)
+    lr.fit(trainingSet)
```
jkbradley: No need to call fit for each alg.  This should be done for all Estimator tests.

src/main/scala/com/databricks/spark/sql/perf/mllib/clustering/LDA.scala
```diff
@@ -0,0 +1,60 @@
+package com.databricks.spark.sql.perf.mllib.clustering
+
+import com.databricks.spark.sql.perf.mllib.{ClassificationContext, TestFromTraining, BenchmarkAlgorithm}
+import com.databricks.spark.sql.perf.mllib.OptionImplicits._
+import org.apache.commons.math3.random.Well19937c
+import org.apache.spark.ml.Transformer
+import org.apache.spark.ml.clustering.{LDA => ML_LDA}
```
jkbradley: Why rename?

src/main/scala/com/databricks/spark/sql/perf/mllib/clustering/LDA.scala
```diff
@@ -0,0 +1,60 @@
...
+    import ctx.extraParams._
+    new ML_LDA()
+        .setK(ldaNumTopics)
+        .setSeed(randomSeed.toLong)
+        .setMaxIter(numIterations)
+        .setOptimizer(ldaOptimizer)
+        .fit(trainingSet)
+  }
+
+  def score(
```
jkbradley: Leave a TODO

src/main/scala/com/databricks/spark/sql/perf/mllib/data/logistic_regression.scala
```diff
@@ -0,0 +1,117 @@
...
+import com.databricks.spark.sql.perf.MLTestParameters
+import org.apache.spark.SparkContext
+import org.apache.spark.ml.linalg.Vectors
+import org.apache.spark.ml.linalg.Vector
+import org.apache.spark.mllib.random._
+import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{SQLContext, DataFrame}
+
+object DataGenerator {
```
jkbradley: Rename file

src/main/scala/com/databricks/spark/sql/perf/mllib/data/logistic_regression.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def generateBinaryLabeledPoints(
```
jkbradley: Remove unnecessary code please

src/main/scala/com/databricks/spark/sql/perf/mllib/yaml.scala
```diff
@@ -0,0 +1,184 @@
...
+
+
+/**
+ * The configuration information generated from reading a YAML file.
+ *
+ * @param output the output direct
+ * @param timeout
+ * @param runnableBenchmarks
+ */
+case class YamlConfig(
```
jkbradley: I'm not going to review this file.  Let's make sure to check the perf results carefully to make sure they are using the correct params.  Also, can you make a note to add unit tests later on?

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -97,3 +102,48 @@ case class BreakdownResult(
...
+
+  val empty = MLTestParameters()
+}
+
+// KEEP ARGUMENTS SORTED BY NAME
+case class ExtraMLTestParameters(
+    ldaDocLength: Option[Int] = None,
+    ldaNumVocabulary: Option[Int] = None,
+    ldaNumTopics: Option[Int] = None,
+    ldaOptimizer: Option[String] = None,
```
jkbradley: For common params, I'd just share 1 param: "optimizer" instead of "ldaOptimizer"

src/main/scala/org/apache/spark/ml/ModelBuilder.scala
```diff
@@ -0,0 +1,17 @@
...
+ */
+object ModelBuilder {
+
+  def newLogisticRegressionModel(
+      coefficients: Vector,
+      intercept: Double): LogisticRegressionModel = {
+    new LogisticRegressionModel("lr", coefficients, intercept)
+  }
+
+}
```
jkbradley: style here and elsewhere

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -97,3 +102,48 @@ case class BreakdownResult(
...
+    numTestExamples: Option[Long] = None,
+    numPartitions: Option[Int] = None,
+    randomSeed: Option[Int] = None)
+
+object MLTestParameters {
+
+  val empty = MLTestParameters()
+}
+
+// KEEP ARGUMENTS SORTED BY NAME
```
jkbradley: Explain why.

jkbradley:
@thunterdb Thanks for the PR.  I did an initial pass.  After the PR is updated, I can do another pass.

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
```diff
@@ -66,25 +66,6 @@ abstract class Benchmark(
...
-   * two options, 200 and 2000. The setup is used to set the value of property
-   * "spark.sql.shuffle.partitions".
-   *
-   * {{{
-   *   Variation("shufflePartitions", Seq("200", "2000")) {
-   *     case num => sqlContext.setConf("spark.sql.shuffle.partitions", num)
-   *   }
-   * }}}
-   */
-  case class Variation[T](name: String, options: Seq[T])(val setup: T => Unit)
```
yhuai: Are we making any actual changes? Or, we are just moving these classes?

thunterdb: No, this is just moving the classes out, otherwise the scala compiler gets
very confused by the path-dependent types when there is an issue elsewhere.

yhuai: Cool

src/main/scala/com/databricks/spark/sql/perf/mllib/MLAlgorithmTests.scala
thunterdb: Yes, sorry, removing this file

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
thunterdb: I was not sure, because `Benchmark` does pass it around as an argument, and it gets overriden in the notebooks.

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
thunterdb: Oh yes!!! Thanks.

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
thunterdb: This is mostly a cosmetic issue and I will be happy to revert it. When we fork. we break the progress reporting in databricks, and we do not have progress bars and other information. Since it is also accessible through the UI, I am tempted to revert this change.

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
thunterdb: a `Duration` object to alleviate your concerns about unit. This can be converted to millis, nanos, etc.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
thunterdb: A downside is that the implementation of the scoring function may have to do a cast, but it should not be much of an issue. Will do.

src/main/scala/com/databricks/spark/sql/perf/mllib/ClassificationContext.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/ClassificationContext.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/MLBenchmarks.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/MLClassificationBenchmarkable.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/MLClassificationBenchmarkable.scala
thunterdb: done for all members

src/main/scala/com/databricks/spark/sql/perf/mllib/MLLib.scala
thunterdb: Done. It also follows the same interface

src/main/scala/com/databricks/spark/sql/perf/mllib/OptionImplicits.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
thunterdb: Moved to the default implementation

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
thunterdb: Making the methods protected.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
thunterdb: This would not let us use transformers then? I suggest we keep returning a transformer for this method instead of a model.

src/main/scala/com/databricks/spark/sql/perf/mllib/clustering/LDA.scala
thunterdb: Because it conflicts with the name of the class below. I am going to remove this import.

src/main/scala/com/databricks/spark/sql/perf/mllib/clustering/LDA.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/data/logistic_regression.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/data/logistic_regression.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/mllib/yaml.scala
thunterdb: Sounds good

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -97,3 +102,48 @@ case class BreakdownResult(
...
+case class ExtraMLTestParameters(
```
thunterdb: You are right, I am going to combine both classes.

Regarding the number of fields, this issue will disappear when we move this year to scala 2.11, and we are already reusing some parameters.

I am not keen on a big map, because the typing information acts as a basic sanity check.

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -77,10 +80,12 @@ case class BenchmarkResult(
     result: Option[Long] = None,
     breakDown: Seq[BreakdownResult] = Nil,
     queryExecution: Option[String] = None,
-    failure: Option[Failure] = None)
+    failure: Option[Failure] = None,
+    ml: Option[MLResult] = None)
```
jkbradley: The test params should be stored separately from the result.  That way, when "ml" is not stored, the test params will still be available in the returned BenchmarkResult.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -0,0 +1,94 @@
...
+      model: Transformer): Double = -1.0 // Not putting NaN because it is not valid JSON.
+}
+
+/**
+ * Uses an evaluator to perform the scoring.
+ */
+trait ScoringWithEvaluator {
+  self: BenchmarkAlgorithm =>
+
+  protected def evaluator(ctx: MLBenchContext): Evaluator = new MulticlassClassificationEvaluator()
```
jkbradley: This is specific to classification, not all ScoringWithEvaluator types.

src/main/scala/com/databricks/spark/sql/perf/mllib/MLBenchContext.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.spark.sql.SQLContext
+
+
+/**
+ * All the information required to run a test.
+ *
+ * @param extraParams
+ * @param sqlContext
+ */
+case class MLBenchContext(
```
jkbradley: If the seed is always stored in MLParams, then can we eliminate this class?

src/main/scala/com/databricks/spark/sql/perf/mllib/MLBenchmarks.scala
```diff
@@ -0,0 +1,37 @@
...
+import com.databricks.spark.sql.perf.mllib.classification.LogisticRegression
+import org.apache.spark.SparkContext
+import org.apache.spark.sql.SQLContext
+
+import com.databricks.spark.sql.perf.{MLParams}
+import OptionImplicits._
+
+case class MLTest(
+    benchmark: BenchmarkAlgorithm,
+    extra: MLParams)
```
jkbradley: Rename "extra" here and elsewhere

src/main/scala/com/databricks/spark/sql/perf/mllib/MLTransformerBenchmarkable.scala
```diff
@@ -0,0 +1,107 @@
+package com.databricks.spark.sql.perf.mllib
+
+import com.databricks.spark.sql.perf._
+import com.typesafe.scalalogging.slf4j.Logging
+import org.apache.spark.sql._
+
+import scala.collection.mutable.ArrayBuffer
+
+class MLTransformerBenchmarkable(
```
jkbradley: Isn't this for Estimator, not Transformer?

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -97,3 +102,43 @@ case class BreakdownResult(
     delta: Double)
 
 case class Failure(className: String, message: String)
+
+// KEEP ARGUMENTS SORTED BY NAME.
+// It simplifies lookup when checking if a parameter is here already.
+case class MLParams(
+    // *** Common to all algorithms ***
+    randomSeed: Option[Int] = None,
```
jkbradley: It would be nice to always store the seed in results.  How about making the default value be Some(42)?

src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
jkbradley: Gotcha.  @yhuai what do you think?

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
jkbradley: Your call : )

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
jkbradley: I don't think it will need to.  For a specific alg, that alg's test should set the output columns and override the Evaluator setup as needed, so the generic transform() and evaluate() calls should handle everything.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/LogisticRegression.scala
jkbradley: You're right: returning a Transformer is better.

src/main/scala/com/databricks/spark/sql/perf/results.scala
jkbradley: SGTM

yhuai:
build.sbt and shared components look good to me.

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: I moved the ml test params to `BenchmarkResult`. Note that the regular sql parameters, which are an empty map in the code base (???), are stored separately in `parameters`.

src/main/scala/com/databricks/spark/sql/perf/mllib/MLTransformerBenchmarkable.scala
thunterdb: This is the mother of all issues with subclassing, and I think the jury is still out on this one. Since Estimator is a transformer with an extra fitting operation, you could argue to either make 2 classes MLTransformerBenchmarkable and MLEstimatorBenchmarkable (the latter even being a subclass of the former), or you could  argue to have one class general MLGeneralTransformerBenchmarkable, for which the the `fit` method is a no-op in case of transformers.

I do not have a strong preference for either pattern, so if you think that class should be restricted to Estimators, I will be happy to do that.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
thunterdb: Moving the default implementation to logistic regression. This is only one line of code that we can factorize later anyway.

src/main/scala/com/databricks/spark/sql/perf/mllib/MLBenchContext.scala
thunterdb: Hmm right. This was meant to use time as a seed when left unspecified, but the seed should always be specificed. I am going to remove that.

thunterdb: That class is still useful I find for passing the sql context and the spark context around, I find.

src/main/scala/com/databricks/spark/sql/perf/mllib/MLBenchmarks.scala
thunterdb: oh, thanks! I checked the whole code base this time.

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: Done

thunterdb:
@jkbradley comments addressed - except for one for which I would like your opinion. 

I also reran the sql tests and got verbal agreement from @yhuai

src/main/scala/com/databricks/spark/sql/perf/mllib/MLTransformerBenchmarkable.scala
jkbradley: No, it's fine, and we can always restructure if needed.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -97,3 +103,42 @@ case class BreakdownResult(
     delta: Double)
 
 case class Failure(className: String, message: String)
+
+// KEEP ARGUMENTS SORTED BY NAME.
+// It simplifies lookup when checking if a parameter is here already.
+case class MLParams(
+    // *** Common to all algorithms ***
+    randomSeed: Option[Int] = Some(0),
```
jkbradley: Set to something other than 0, or check that this is OK.  I recall running into some RNG in Spark which always generated 0 whenever given a seed of 0.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -97,3 +103,42 @@ case class BreakdownResult(
...
+    numExamples: Option[Long] = None,
+    numTestExamples: Option[Long] = None,
+    numPartitions: Option[Int] = None,
+    // *** Specialized and sorted by name ***
+    ldaDocLength: Option[Int] = None,
+    ldaNumVocabulary: Option[Int] = None,
```
jkbradley: Can you please make the param names match those in MLlib?  It will making writing configs easier.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -77,10 +80,13 @@ case class BenchmarkResult(
     result: Option[Long] = None,
     breakDown: Seq[BreakdownResult] = Nil,
     queryExecution: Option[String] = None,
-    failure: Option[Failure] = None)
+    failure: Option[Failure] = None,
+    mlParameters: Option[MLParams] = None,
+    ml: Option[MLResult] = None)
```
jkbradley: Rename to mlResult

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -77,10 +80,13 @@ case class BenchmarkResult(
     result: Option[Long] = None,
     breakDown: Seq[BreakdownResult] = Nil,
     queryExecution: Option[String] = None,
-    failure: Option[Failure] = None)
+    failure: Option[Failure] = None,
+    mlParameters: Option[MLParams] = None,
```
jkbradley: rename to MLParams to be consistent

jkbradley:
I just had a few comments.  Other than those, LGTM.  Feel free to merge this after updating those.  Thanks!

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: Good point, done

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: Yes, done.

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: Done

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: Done

thunterdb:
@jkbradley thanks for the review, merging.

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: (the remaining ones are specific to the data generation procedure)

yhuai:
merging!

PR #72, opened 2016-06-27
Generalized linear models performance tests

thunterdb:
This adds GLM regression to the tests, and fixes some issues that were not addressed before in 0.4.9 mllib tests.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
@@ -45,9 +45,9 @@ trait Benchmarkable extends Logging {
     sparkContext.setJobDescription(s"Execution: $name, $description")
     beforeBenchmark()
     val result = if (forkThread) {
-      runBenchmarkForked(includeBreakdown, description, messages, timeout)
-    } else {
       doBenchmark(includeBreakdown, description, messages)
+    } else {
+      runBenchmarkForked(includeBreakdown, description, messages, timeout)
     }
     afterBenchmark(sqlContext.sparkContext)
     result
@@ -81,7 +81,13 @@ trait Benchmarkable extends Logging {
         } catch {
           case e: Throwable =>
             logger.info(s"$that: failure in runBenchmark: $e")
-            throw e
+            println(s"$that: failure in runBenchmark: $e")
+            result = BenchmarkResult(
+              name = name,
+              mode = executionMode.toString,
+              parameters = Map.empty,
+              failure = Some(Failure(e.getClass.getSimpleName,
+                e.getMessage + ":\n" + e.getStackTraceString)))
         }
       }
     }
--- a/src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
@@ -45,9 +45,9 @@ trait Benchmarkable extends Logging {
     sparkContext.setJobDescription(s"Execution: $name, $description")
     beforeBenchmark()
     val result = if (forkThread) {
-      doBenchmark(includeBreakdown, description, messages)
-    } else {
       runBenchmarkForked(includeBreakdown, description, messages, timeout)
+    } else {
+      doBenchmark(includeBreakdown, description, messages)
     }
     afterBenchmark(sqlContext.sparkContext)
     result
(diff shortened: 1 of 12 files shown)


src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
```diff
@@ -45,9 +45,9 @@ trait Benchmarkable extends Logging {
     sparkContext.setJobDescription(s"Execution: $name, $description")
     beforeBenchmark()
     val result = if (forkThread) {
-      runBenchmarkForked(includeBreakdown, description, messages, timeout)
-    } else {
       doBenchmark(includeBreakdown, description, messages)
+    } else {
```
thunterdb: that issue had not been fixed

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
```diff
@@ -81,7 +81,13 @@ trait Benchmarkable extends Logging {
         } catch {
           case e: Throwable =>
             logger.info(s"$that: failure in runBenchmark: $e")
-            throw e
+            println(s"$that: failure in runBenchmark: $e")
```
thunterdb: also printing on the standard output

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
jkbradley: Are you sure?  Isn't this reverting the fix?

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -82,7 +90,9 @@ trait TestFromTraining {
     // Copy the context with a new seed.
     val ctx2 = ctx.params.randomSeed match {
       case Some(x) =>
-        val p = ctx.params.copy(randomSeed = Some(x + 1))
+        // Also set the number of examples to the number of test examples.
```
jkbradley: This is odd; it makes me think passing a "context" to trainingDataSet() is not useful.  It would be simpler to pass just what is needed: the number of examples, SQLContext, etc.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -113,9 +113,12 @@ case class MLParams(
     numTestExamples: Option[Long] = None,
     numPartitions: Option[Int] = None,
     // *** Specialized and sorted by name ***
+    elasticNetParam: Option[Double] = None,
+    family: Option[String] = None,
+    link: Option[String] = None,
+    k: Option[Int] = None,
```
jkbradley: @yhuai What are restrictions on changing the order of arguments for case classes used within BenchmarkResult?  Will it break reading from existing log files using the previous schema?

Maybe it's not a problem if the last commit's version was not yet deployed though.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.4.9-SNAPSHOT"
+version in ThisBuild := "0.4.10-SNAPSHOT"
```
jkbradley: Not sure myself: What's the convention here?  Is it incremented when the result schema changes?

src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
thunterdb: You are right. It works on databricks, though, so there must be an error somewhere else.

thunterdb: Found the real bug.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
thunterdb: Yes, I tried to do this at first, but it turns out it is useful when building the training set from a 'true' model.

src/main/scala/com/databricks/spark/sql/perf/results.scala
thunterdb: I talked with @brkyvz and he said it is fine

version.sbt
thunterdb: Since it is changing schemas, I would rather bump the version, but there is no rule for this package as far as I know. In any case, we are not following semantic versioning!

thunterdb:
@jkbradley comments addressed. Let me know if you have any questions.

jkbradley:
LGTM, merging with master
Thanks!

PR #75, opened 2016-07-02
Added kmeans test

jkbradley:
Ran tests locally, and they worked.  Also fixed missing param in LogReg test in small yaml file

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/data/dataGeneration.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/data/dataGeneration.scala
@@ -36,6 +36,23 @@ object DataGenerator {
       new FeaturesGenerator(featureArity), numExamples, numPartitions, seed)
     sql.createDataFrame(rdd.map(Tuple1.apply)).toDF("features")
   }
+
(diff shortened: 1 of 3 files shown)


jkbradley:
CC: @thunterdb

src/main/scala/com/databricks/spark/sql/perf/mllib/data/dataGeneration.scala
```diff
@@ -78,3 +95,46 @@ class FeaturesGenerator(val featureArity: Array[Int])
...
+    }
+  }
+
+  private val centers = (0 until numCenters).map{i =>
+    Array.fill(numFeatures)((2 * rng.nextDouble() - 1) * scale_factors(i))
+  }
+
+  override def nextValue(): Vector = {
+    val pick_center_rand = rng2.nextDouble()
+    val center = centers(concentrations.indexWhere(p => pick_center_rand <= p))
```
thunterdb: this is correct and we do not care too much about performance, but I think that apache commons has an implementation.

thunterdb:
Just a small comment that is not super important, merging.

PR #76, opened 2016-07-05
ALS algorithm for spark-sql-perf

thunterdb:
This has been tested locally with a small amount of data.

I have not bothered to reimplement a more robust version of the ALS synthetic data generation, so it will still require some manual parameter tweaking as before.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/data/dataGeneration.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/data/dataGeneration.scala
@@ -2,8 +2,9 @@ package com.databricks.spark.sql.perf.mllib.data
 
 import org.apache.spark.ml.linalg.Vectors
 import org.apache.spark.ml.linalg.Vector
+import org.apache.spark.ml.recommendation.ALS.Rating
 import org.apache.spark.mllib.random._
-import org.apache.spark.rdd.RDD
+import org.apache.spark.rdd.{PairRDDFunctions, RDD}
 import org.apache.spark.sql.{SQLContext, DataFrame}
 
 
@@ -23,6 +24,7 @@ object DataGenerator {
 
   /**
    * Generate a mix of continuous and categorical features.
+ *
    * @param featureArity  Array of length numFeatures, where 0 indicates a continuous feature and
    *                      a value > 0 indicates a categorical feature with that arity.
    */
@@ -39,6 +41,7 @@ object DataGenerator {
 
   /**
    * Generate data from a Gaussian mixture model.
+ *
    * @param numCenters  Number of clusters in mixture
    */
   def generateGaussianMixtureData(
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/sql/perf/mllib/recommendation/ALS.scala
```diff
@@ -0,0 +1,59 @@
...
+      .setRank(rank)
+      .setMaxIter(maxIter)
+  }
+
+  override def score(
+      ctx: MLBenchContext,
+      testSet: DataFrame,
+      model: Transformer): Double = {
+    val out = model.transform(testSet)
+    val prediction = out("prediction")
```
thunterdb: I rewrote this part to implement RMSE using column operations, which seems more readable than the old RDD-based operations.

src/main/scala/com/databricks/spark/sql/perf/mllib/data/dataGeneration.scala
```diff
@@ -53,11 +54,60 @@ object DataGenerator {
       seed)
     sql.createDataFrame(rdd.map(Tuple1.apply)).toDF("features")
   }
+
+  def generateRatings(
+      sql: SQLContext,
+      numUsers: Int,
+      numProducts: Int,
+      numRatings: Long,
```
jkbradley: How about taking numTestRatings so it's no longer hard-coded?  This will let you generate the right # of test examples below.

src/main/scala/com/databricks/spark/sql/perf/mllib/recommendation/ALS.scala
```diff
@@ -0,0 +1,59 @@
...
+      ctx.seed())._1
+  }
+
+  override def testDataSet(ctx: MLBenchContext): DataFrame = {
+    import ctx.params._
+    DataGenerator.generateRatings(
+      ctx.sqlContext,
+      numUsers,
+      numItems,
+      numTestExamples,
```
jkbradley: This won't be the right number of ratings.

src/main/scala/com/databricks/spark/sql/perf/mllib/recommendation/ALS.scala
```diff
@@ -0,0 +1,59 @@
...
+    import ctx.params._
+    new ml.recommendation.ALS()
+      .setSeed(ctx.seed())
+      .setRegParam(regParam)
+      .setNumBlocks(numPartitions)
+      .setRank(rank)
+      .setMaxIter(maxIter)
+  }
+
+  override def score(
```
jkbradley: Why not use ScoringWithEvaluator with RegressionEvaluator?

jkbradley:
That's it!

src/main/scala/com/databricks/spark/sql/perf/mllib/recommendation/ALS.scala
thunterdb: oh yes, good point

thunterdb:
@jkbradley comments addressed, I am going to merge

src/main/scala/com/databricks/spark/sql/perf/mllib/recommendation/ALS.scala
thunterdb: Good point, done

PR #77, opened 2016-07-05
Linear regression

thunterdb:
Adds a specific check for linear regression with the L-BFGS solver.

Tested on a small amount of data.

### DIFF
--- a/src/main/scala/configs/mllib-small.yaml
+++ b/src/main/scala/configs/mllib-small.yaml
@@ -65,6 +65,8 @@ benchmarks:
       maxIter: 3
   - name: regression.LinearRegression
     params:
+      numExamples: 100
+      numTestExamples: 100
       numFeatures: 100
       regParam: 0.1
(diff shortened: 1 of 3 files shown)


src/main/scala/configs/mllib-small.yaml
```diff
@@ -63,3 +63,9 @@ benchmarks:
       numClasses: 4
       numFeatures: 5
       maxIter: 3
+  - name: regression.LinearRegression
+    params:
+      numFeatures: 100
```
jkbradley: numExamples, numTestExamples?

jkbradley:
Just the 1 comment

src/main/scala/configs/mllib-small.yaml
thunterdb: Added

thunterdb:
Thanks, merging

jkbradley:
LGTM, you can merge

PR #82, opened 2016-07-19
Removes labels from tree data generation

thunterdb:
Note: it builds upon #81 , which should be reviewed first.

### DIFF
--- a/src/main/scala/configs/mllib-small.yaml
+++ b/src/main/scala/configs/mllib-small.yaml
@@ -60,7 +60,7 @@ benchmarks:
       numExamples: 100
       numTestExamples: 10
       depth: 3
-      numClasses: 4
+      numClasses: 2
       numFeatures: 5
       maxIter: 3
   - name: regression.LinearRegression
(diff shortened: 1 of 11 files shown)


thunterdb:
It actually needs to be added at a different place. WIP.

jkbradley:
LGTM, thanks for realizing this.  Not sure why tests seemed to pass locally.  Maybe it's just unclear to me how to check for whether things pass; is that documented anywhere, or is it just buried in logs which get deleted after the tests run?  (I recall I had to hack the test to make it die and not delete the logs before exiting.)

src/main/scala/configs/mllib-small.yaml
```diff
@@ -60,7 +60,7 @@ benchmarks:
       numExamples: 100
       numTestExamples: 10
       depth: 3
-      numClasses: 4
+      numClasses: 2
```
thunterdb: GBT only handles binary features, it seems

jkbradley: Yep

jkbradley:
Now it LGTM!  Though I have not tested it myself

PR #90, opened 2016-09-30
[SC-4409] Correct query 41 in TPCDS kit

srinathshankar:
This query should have a predicate of the form
```
((a=b) and P1) 
or 
((a=b) and P2)
```
It used to fail since we were previously not able to extract the common predicate `(a=b)` and rewrite as
```
(a=b) and (P1 or P2)
```
This defect was fixed with https://issues.apache.org/jira/browse/SPARK-15122
The manual rewrite currently checked in is however `(a=b) and P1 or P2` which is not the same as the original (and which spark cannot run). 
This patch restores the original version of the query.


srinathshankar:
cc @davies @JoshRosen

srinathshankar:
cc @hvanhovell

srinathshankar:
@sameeragarwal

sameeragarwal:
LGTM, merging!

PR #96, opened 2016-12-08
Show std dev as a percentage, show all queries

a-roberts:
Add an extra column to the table that provides the standard deviation as a percentage (like all other results in our table this should be improved by formatting to two decimal places).

Also call .show(100) so by default we'll print all results instead of just the first 20.


npoggi:
@a-roberts could you rebase this commit and add me as the reviewer to finally getting this PR merged?  Thanks

PR #98, opened 2017-03-15
Add option to avoid cleaning after each run, to enable parallel runs

ericl:
Without this change, parallel TPCDS runs on a cluster will sometimes fail with cryptic broadcast failed errors.


ericl:
Made a new release. Plus, we should default to Scala 2.11

JoshRosen:
Makes sense; thanks!

PR #101, opened 2017-05-12
Add tpcds 2.4 queries

vlyubin:
Adds 2.4 version of the queries in addition to 1.4 that's already there.

Tests:

Made a jar and used it to run a new benchmark.


src/main/scala/com/databricks/spark/sql/perf/tpcds/TPCDS_2_4_Queries.scala
```diff
@@ -0,0 +1,3839 @@
...
+ * (noted for each query).
+ * Don't modify this except for these kind of modifications.
+ */
+trait Tpcds_2_4_Queries extends Benchmark {
+
+  import ExecutionMode._
+
+  // Queries the TPCDS 2.3 queries using the qualifcations values in the templates.
+  val tpcds2_4Queries = Seq(
+      ("q1", """
```
sameeragarwal: nit: How about we move each query into a separate file for better readability? Here's how we did this in Spark: https://github.com/apache/spark/tree/master/sql/core/src/test/resources/tpcds/

rxin: yea it'd be great to put queries in files ...

vlyubin: Ok, did that, retested on Databricks

src/main/resources/tpcds_2_4/q23a.sql
```diff
@@ -0,0 +1,43 @@
...
+        where ss_customer_sk = c_customer_sk
+         and ss_sold_date_sk = d_date_sk
+         and d_year in (2000, 2000+1, 2000+2,2000+3)
+        group by c_customer_sk)),
+ best_ss_customer as
+ (select c_customer_sk,sum(ss_quantity*ss_sales_price) ssales
+  from store_sales, customer
+  where ss_customer_sk = c_customer_sk
+  group by c_customer_sk
+  having sum(ss_quantity*ss_sales_price) > (50/100.0) *
```
juliuszsompolski: Note: Official query template has 95/100, and it seems that it was our bug to use 50/100.
See e.g. https://github.com/gregrahn/tpcds-kit/blob/master/query_templates/query23.tpl and https://github.com/pivotalguru/TPC-DS/blob/master/05_sql/123.imp.23.sql

src/main/resources/tpcds_2_4/q23b.sql
```diff
@@ -0,0 +1,50 @@
...
+         and ss_sold_date_sk = d_date_sk
+         and d_year in (2000, 2000+1, 2000+2,2000+3)
+        group by c_customer_sk) x),
+ best_ss_customer as
+ (select c_customer_sk,sum(ss_quantity*ss_sales_price) ssales
+  from store_sales
+      ,customer
+  where ss_customer_sk = c_customer_sk
+  group by c_customer_sk
+  having sum(ss_quantity*ss_sales_price) > (50/100.0) *
```
juliuszsompolski: Ditto 95/100

src/main/resources/tpcds_2_4/q23a.sql
vlyubin: kk, will fix

src/main/resources/tpcds_2_4/q23b.sql
gregrahn: Depends on what you want.  The Qualification Substitution Parameters for query23 are in Appendix B in the TPC-DS spec.

Qualification Substitution Parameters for query23.tpl:
* MONTH.01 = 2
* YEAR.01 = 2000
* TOPPERCENT=50

juliuszsompolski: Hm. Indeed. On the other hand, the text just two lines above that says "Compute the best store customers as those that are in the 5th percentile of sales", and the query23.tpl in the TPC DS kit downloaded from tpc.org (and the copy from your repo) has ` define TOPPERCENT=random(95,95,uniform); `
...
but that wouldn't be the only discrepancy between the text query description in Appendix B would be inconsistent with the query template in query23.tpl anyway...

gregrahn: The mismatch between the substitution variables and the English question is a red herring.  There are answer sets for qualification queries so it's easy to validate what is correct.

I come back to understanding the goal here.  There really should be just two forms of queries: 

- The 1GB qualification version with substitution variables from the spec.  Presto uses these in their tests ([link](https://github.com/prestodb/presto/tree/master/presto-benchto-benchmarks/src/main/resources/sql/presto/tpcds))
- Query templates that are used to generate query files for a given data scale factor.  Different substitution variables are used for difference data sizes so it's technically incorrect to use the templates to generate query files for 1GB but then use those against 10TB.

vlyubin:
Merging as this was approved.

kiszk:
@rxin @sameeragarwal Is it better to put these files for `TPCDSQueryBenchmark`, too?

PR #104, opened 2017-05-25
Rewrite Readme; make compatible with TPC-DS 2.4

dzhang-bap:
I recently tried to use the tool to conduct some TPC-DS analysis, and it was quite a journey, largely because the instructions were seriously lacking. David DeWitt suggested that I contribute back to Databricks. Srinath suggested that I create a pull request. So there you are.

In case interested, below was my journey.

1.       The first command at the perf tool’s manual https://github.com/databricks/spark-sql-perf is “bin/run --help”.
It tried to reach out to the network but our dev server was behind a firewall. I have set HTTP proxy which worked for other programs but this script complained about failing to parse http_proxy and https_proxy.
After I manually set the environmental variables, the script runs but complained it cannot get sbt. I installed sbt and added it to path. But the script was still not happy.
Resolution: I gave up dev server, and ran it on my Mac laptop.
 
2.       After I finished the “quick start” section, I was at a total loss what to do next.
The manual says dsdgen is needed. I got one from tpc.org. It did not support building on Mac but I made it build. This part was not too hard.
The next paragraph in the manual is a pseudo-code block.
Where am I supposed to type the pseudo code? What’s the relationship between the pseudo code and the “bin/run” script? What’s the relationship between the pseudo code and my Spark cluster?
Resolution: I spent a lot of time studying various documents and source code, and eventually figured out all these. In particular, I figured that I should start my spark cluster using “spark-shell” while providing it a “driver-class-path” parameter specifying the folder that was created by the “bin/run” script.
 
3.       I had no clue what to use for variables mentioned in the code block.
For instance, in
val tables = new Tables(sqlContext, dsdgenDir, scaleFactor)
What should I use for sqlContext?
In
tables.genData(location, format, overwrite, partitionTables, useDoubleForDecimal, clusterByPartitionColumns, filterOutNullPartitionValues)
What should I use for location, format, and other variables?
Resolution: I carefully studied the source code, and did a lot of trials-and-errors to figure out.
 
4.       Incompatible versions of dsdgen.
I ran the tables.genData(…) command from the code block. It generated some files, but all files included an error message “option filter or its argument unknown”.
Resolution: I figured that this was due to a version mismatch: the perf tool supports an older version of TPC-DS tool. I modified the perf tool source code to work with the newer version of dsdgen.
 
5.       Avoiding repeated calls to dsdgen.
After I made tables.genData(…) successfully call dsdgen, I saw 24 dat files generated (good!), but no data was loaded into Spark’s parquet file folder. To debug the problem I had to iteratively change something and then call tables.genData(…). In each iteration I had to wait for the same 24 dat files to be generated. That was wasteful.
Resolution: I modified the perf tool code to reuse the already generated dat files if exist.
 
6.       Wrong encoding.
Nine out of 24 dat files were loaded into the Spark tables, before an exception was thrown. The exception was java.nio.charset.MalformedInputException.
Resolution: I did some google search and realized that when opening the dat files I should use the “Cp1252” encoding. Now all 24 dat files were loaded.
 
7.       Failure to create a TPCDS object.
The next step in the pseudo code is to create a TPCDS object. It failed – in fact at this point I was sort of expecting it to fail, because all previous steps failed upon first try. The exception was missing dependencies.
Resolution: As I was no longer enthusiastic about debugging the perf tool, I decided to cut corners. I already figured how to generate query plans given a query string. I decided to just use the 104 query strings from TPCDS class, instead of generating a TPCDS object.
 
8.       Last hurdle: not all queries work.
I wrote a script that prints out the query plans for each of the 104 query strings. It successfully generated plans but only for 28 queries. Debugging revealed that the parsing of the 29th query threw an exception.
Resolution: I modified my script to surround each parsing statement with a try-catch block, to replace the exception with an error message so that the rest of the queries have a chance to be parsed. I found 99 out of 104 queries were able to be parsed. The five failed queries had an error saying inner joins are missing non-trivial join condition.
 
Note that this pull request did not address the dependency issue at step 7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,59 +2,111 @@
 
 [![Build Status](https://travis-ci.org/databricks/spark-sql-perf.svg)](https://travis-ci.org/databricks/spark-sql-perf)
 
-This is a performance testing framework for [Spark SQL](https://spark.apache.org/sql/) in [Apache Spark](https://spark.apache.org/) 1.6+.
+This is a performance testing framework for [Spark SQL](https://spark.apache.org/sql/) in [Apache Spark](https://spark.apache.org/) 1.6+. The toolkit includes the TPC-DS benchmark and a few other sample benchmarks.
 
-**Note: This README is still under development. Please also check our source code for more information.**
+# Build
 
-# Quick Start
+Suppose the spark-sql-perf tool is located at ~/spark-sql-perf. The (non-intuitive) command to build the benchmark is:
 
 ```
-$ bin/run --help
-
-spark-sql-perf 0.2.0
-Usage: spark-sql-perf [options]
-
-  -b <value> | --benchmark <value>
-        the name of the benchmark to run
-  -f <value> | --filter <value>
-        a filter on the name of the queries to run
-  -i <value> | --iterations <value>
-        the number of iterations to run
-  --help
-        prints this usage text
-        
-$ bin/run --benchmark DatasetPerformance
+cd ~/spark-sql-perf
+bin/run --help
 ```
 
-# TPC-DS
+The first time you run the command, it may take a while because it may download and install dependencies such as sbt. Future runs should be faster.
 
-## How to use it
-The rest of document will use TPC-DS benchmark as an example. We will add contents to explain how to use other benchmarks add the support of a new benchmark dataset in future.
+# Run a Benchmark
 
-### Setup a benchmark
-Before running any query, a dataset needs to be setup by creating a `Benchmark` object. Generating
-the TPCDS data requires dsdgen built and available on the machines. We have a fork of dsdgen that
-you will need. It can be found [here](https://github.com/davies/tpcds-kit).  
+Have a look at src/main/scala/com/databricks/spark/sql/perf/DatasetPerformance.scala. It defines a class `DatasetPerformance` which derives from Benchmark. To run this particular benchmark:
+
+```
+cd ~/spark-sql-perf
+bin/run --benchmark DatasetPerformance
+```
+
+# The TPC-DS Benchmark
+
+## Build dsdgen
+
+This step builds a data generation tool that will be called by `tables.genData()` in a later step.
+Download TPC-DS Tools from http://tpc.org.
+Another option is to use Databricks' fork of dsdgen [here](https://github.com/davies/tpcds-kit).
+
+Suppose the downloaded zip file is ~/tpc_ds/127d3fa4-fac9-43f7-b1c5-6a00c2291ab9-tpc-ds-tool.zip, and the version is v2.4.0. To unzip and build:
+
+```
+cd ~/tpc_ds
+unzip 127d3fa4-fac9-43f7-b1c5-6a00c2291ab9-tpc-ds-tool.zip
+cd v2.4.0/tools
+make
+```
+
+If the build is successful, you should get an executable `dsdgen`.
+
+**Note:** It may be necessary to change `ReportErrorNoLine(..., 1)` to `ReportErrorNoLine(..., 0)` in line 492 of driver.c, to avoid a error when running `tables.genData()` at a later step with error message like "Table ... is a child; it is populated during the build of its parent (e.g., catalog_sales builds catalog returns)."
+
+**Note for Mac:** The dsdgen tool does not build on Mac. But you can easily make it build. Essentially, you should include `stdlib.h` instead of `malloc.h`, include `limits.h` instead of `values.h`, and `#define MAXINT INT_MAX`.
+
+## Build Spark
+
+You may get Spark from http://github.com/apache/spark.
+
+Suppose the spark directory is ~/spark. To build:
+
+```
+cd ~/spark
+build/mvn -DskipTests -T 8C -Dskip -Phive -Phive-thriftserver package
+```
+
+In the above command, `-T 8C` means to use 8 threads to build; `-Phive -Phive-thriftserver` means to build with Hive support.
+
+## Start a spark-shell with access to spark-sql-perf classes
+
+Supose the previous step that builds spark-sql-perf generated a folder scala-2.11 in ~/spark-sql-perf/target.
+
+```
+cd ~/spark
+bin/spark-shell --driver-class-path ~/spark-sql-perf/target/scala-2.11/classes
+```
+
+## Generate data, create tables, and create a TPCDS object
+
+Run the following commands in the spark-shell.
 
 ```
 import com.databricks.spark.sql.perf.tpcds.Tables
-// Tables in TPC-DS benchmark used by experiments.
-// dsdgenDir is the location of dsdgen tool installed in your machines.
-val tables = new Tables(sqlContext, dsdgenDir, scaleFactor)
-// Generate data.
-tables.genData(location, format, overwrite, partitionTables, useDoubleForDecimal, clusterByPartitionColumns, filterOutNullPartitionValues)
-// Create metastore tables in a specified database for your data.
-// Once tables are created, the current database will be switched to the specified database.
-tables.createExternalTables(location, format, databaseName, overwrite)
-// Or, if you want to create temporary tables
-tables.createTemporaryTables(location, format)
-// Setup TPC-DS experiment
+import org.apache.spark.SparkContext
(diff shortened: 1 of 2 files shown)


juliuszsompolski:
> Resolution: I modified my script to surround each parsing statement with a try-catch block, to replace the exception with an error message so that the rest of the queries have a chance to be parsed. I found 99 out of 104 queries were able to be parsed. The five failed queries had an error saying inner joins are missing non-trivial join condition.

https://github.com/databricks/spark-sql-perf/pull/101 adds queries from tpcds 2.4, and also fixes this issue with cross joins.

As a workaround with the earlier queries you could have used:
spark.sql.crossJoin.enabled=true

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
```diff
@@ -55,13 +56,21 @@ class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int) extend
             sys.error(s"Could not find dsdgen at $dsdgen or /$dsdgen. Run install")
           }
 
-          // Note: RNGSEED is the RNG seed used by the data generator. Right now, it is fixed to 100.
-          val parallel = if (partitions > 1) s"-parallel $partitions -child $i" else ""
-          val commands = Seq(
-            "bash", "-c",
-            s"cd $localToolsDir && ./dsdgen -table $name -filter Y -scale $scaleFactor -RNGSEED 100 $parallel")
```
juliuszsompolski: The filter option comes from a fork of dsgen in https://github.com/davies/tpcds-kit

gregrahn:
Just as an FYI, it may be easier to integrate the java version of dsdgen found here:
https://github.com/teradata/tpcds

juliuszsompolski:
@gregrahn 
Thanks! I'll take a look at the Teradata's version. For now I brought the fork from your repo up to date at https://github.com/databricks/tpcds-kit

gregrahn:
@juliuszsompolski - As a heads up the `-filter` flag was an unfinished hack that used stdout vs files, but only did so for the `_sales` tables, not the `_returns`.  This hack was reverted when I upgraded to v2.4.  The problem created by this hack was there was no way to generate `_returns` on their own.  See also https://github.com/gregrahn/tpcds-kit/pull/29

juliuszsompolski:
@gregrahn 
I believe @davies has finished this hack in https://github.com/databricks/tpcds-kit/commit/90b8e13a96577cab6788ba644fc0c12642132f18
At least, it seems to work.

gatorsmile:
@juliuszsompolski Maybe we can first merge this PR and then resolve the issue you raised?

juliuszsompolski:
@gatorsmile 
The code change in src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala actually breaks the way it is intended to use, with data generated on the fly with the -filter option using our fork of TPCDS kit in https://github.com/databricks/tpcds-kit/, so that part cannot be merged as-is. Using that fork should be mentioned clearly in the readme to avoid more future confusion...
If we want to support loading from pre-generated files (and not using our fork), that needs to be added as an extra option, not to replace the current behaviour.
Also, the readme changes got a bit outdated after changes merged in #105 and #106.

So it needs some changes before merging.

README.md
```diff
@@ -2,73 +2,126 @@
...
+```
+cd ~/tpc_ds
+unzip 127d3fa4-fac9-43f7-b1c5-6a00c2291ab9-tpc-ds-tool.zip
+cd v2.4.0/tools
+make
+```
+
+If the build is successful, you should get an executable `dsdgen`.
+
+**Note:** It may be necessary to change `ReportErrorNoLine(..., 1)` to `ReportErrorNoLine(..., 0)` in line 492 of driver.c, to avoid a error when running `tables.genData()` at a later step with error message like "Table ... is a child; it is populated during the build of its parent (e.g., catalog_sales builds catalog returns)."
```
juliuszsompolski: Should not be applicable anymore with https://github.com/databricks/tpcds-kit

README.md
```diff
@@ -2,73 +2,126 @@
...
+bin/run --benchmark DatasetPerformance
+```
+
+# The TPC-DS Benchmark
+
+## Build dsdgen
+
+This step builds a data generation tool that will be called by `tables.genData()` in a later step.
+Download TPC-DS Tools from http://tpc.org.
+Another option is to use Databricks' fork of dsdgen [here](https://github.com/davies/tpcds-kit).
```
juliuszsompolski: It is now https://github.com/databricks/tpcds-kit.
It should be the encouraged default by the README, since it provides the correct implementation for the -filter option that helps generate data on the fly.

README.md
```diff
@@ -2,73 +2,126 @@
...
+Suppose the downloaded zip file is ~/tpc_ds/127d3fa4-fac9-43f7-b1c5-6a00c2291ab9-tpc-ds-tool.zip, and the version is v2.4.0. To unzip and build:
+
+```
+cd ~/tpc_ds
+unzip 127d3fa4-fac9-43f7-b1c5-6a00c2291ab9-tpc-ds-tool.zip
+cd v2.4.0/tools
+make
+```
+
+If the build is successful, you should get an executable `dsdgen`.
```
juliuszsompolski: Note: if running on a cluster, dsdgen needs to be available on all worker nodes, because data generation tasks will be running on all workers...
Probably not encountered when running out of local laptop.

README.md
```diff
@@ -2,73 +2,126 @@
...
+// Call dsdgen to create .dat files (if not exists already) in ~/tpc_ds/v2.4.0/tools.
+// Then load into parquet files to be stored at "~/datafiles".
+// The five boolean parameters are:
+// - true  = overwrite
+// - false = partitionTables
+// - true  = useDoubleForDecimal
+// - false = clusterByPartitionColumns
+// - false = filterOutNullPartitionValues
+// See the source code src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala.
+tables.genData("~/datafiles", "parquet", true, false, true, false, false)
```
juliuszsompolski: needs update after #105
also, it was more readable to use descriptive val names than a "true, false, true, false, false" and a comment description.

README.md
```diff
@@ -2,73 +2,126 @@
...
-tables.createTemporaryTables(location, format)
-// Setup TPC-DS experiment
+import org.apache.spark.SparkContext
+import org.apache.spark.sql.SQLContext
+
+val sqlContext = SQLContext.getOrCreate(sc)
+
+// Create a Tables object without generating any data yet.
+// Record the locatin of dsdgen, and scale factor 1.
+val tables = new Tables(sqlContext, "~/tpc_ds/v2.4.0/tools", 1)
```
juliuszsompolski: Outdated and needs update after #106

README.md
```diff
@@ -2,73 +2,126 @@
...
+// - true  = useDoubleForDecimal
+// - false = clusterByPartitionColumns
+// - false = filterOutNullPartitionValues
+// See the source code src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala.
+tables.genData("~/datafiles", "parquet", true, false, true, false, false)
+
+// Generate tables using the parquet files at ~/datafiles, in a database "tpcds".
+// The last parameter means to overwrite.
+// Note that you could also create temporary tables. Again see the source code.
+tables.createExternalTables("~/datafiles", "parquet", "tpcds", true)
```
juliuszsompolski: needs update after #106 + add an example of tables.analyzeTables.

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
juliuszsompolski: Removing the -filter option breaks the way it is intended to be used with https://github.com/databricks/tpcds-kit.

If one wants an option to load from files instead of generating with -filter on the fly, it needs to be implemented as a (non-default) option, not replace the current behaviour.

juliuszsompolski:
I incorporated parts of the readme changes in #120. I add there also Databricks notebooks for running the TPCDS benchmark, which should serve as a better example of how it can be ran manually also.

PR #105, opened 2017-06-01
Add an option to replace Date with String type.

juliuszsompolski:
Spark uses an old fork of Hive, and currently fails with
```
Error in SQL statement: AnalysisException: org.apache.hadoop.hive.ql.metadata.HiveException: java.lang.UnsupportedOperationException: Parquet does not support date. See HIVE-6384;
```
Provide an option to replace Date with String.


juliuszsompolski:
Crap, wrong push...

juliuszsompolski:
databricks:master was forced pushed back and it should be recovered.

PR #106, opened 2017-06-02
Tweaks and improvements

juliuszsompolski:
Data generation:
* Add an option to change Dates to Strings, and specify it in Tables object creator.
* Add discovering partitions to createExternalTables
* Add analyzeTables function that gathers statistics.

Benchmark execution:
* Perform collect() on Dataframe, so that it is recorded by SQL SparkUI.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
@@ -48,7 +48,6 @@ class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int,
      */
     def df(convertToSchema: Boolean, numPartition: Int, filter: Boolean) = {
       val partitions = if (partitionColumns.isEmpty) 1 else numPartition
-      val filterArg = if (filter) "-filter Y" else ""
       val generatedData = {
         sparkContext.parallelize(1 to partitions, partitions).flatMap { i =>
           val localToolsDir = if (new java.io.File(dsdgen).exists) {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
```diff
@@ -43,7 +46,7 @@ class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int) extend
      *  If convertToSchema is true, the data from generator will be parsed into columns and
      *  converted to `schema`. Otherwise, it just outputs the raw data (as a single STRING column).
      */
-    def df(convertToSchema: Boolean, numPartition: Int) = {
+    def df(convertToSchema: Boolean, numPartition: Int, filter: Boolean) = {
```
vlyubin: `filter` seems to be unused here

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
```diff
@@ -257,6 +276,17 @@ class Tables(sqlContext: SQLContext, dsdgenDir: String, scaleFactor: Int) extend
     }
   }
 
+  def analyzeTables(databaseName: String, analyzeColumns: Boolean = false, tableFilter: String = ""): Unit = {
```
vlyubin: minor nit: tableFilter should have been an Option[String], but since it looks like it was a string in `createExternalTables` I guess we can leave it as is ...

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
juliuszsompolski: thx, remnant of an earlier change.

src/main/scala/com/databricks/spark/sql/perf/tpcds/Tables.scala
juliuszsompolski: yeah, wanted to keep it consistent.

vlyubin:
LGTM

PR #110, opened 2017-08-11
Update tests to run with Spark 2.2, add NaiveBayes & Bucketizer ML tests

smurching:
* Made small updates in Benchmark.scala and Query.scala for Spark 2.2
* Added tests for NaiveBayesModel and Bucketizer
* Changed BenchmarkAlgorithm.getEstimator() -> BenchmarkAlgorithm.getPipelineStage() to allow for the benchmarking of Estimators and Transformers instead of just Estimators


src/main/scala/org/apache/spark/ml/NaiveBayesBuilder.scala
```diff
@@ -0,0 +1,16 @@
...
+/** Object for building Naive Bayes classifiers. */
+object NaiveBayesBuilder {
+  def newNaiveBayesModel(pi: Vector, theta: Matrix): NaiveBayesModel = {
+    val model = new NaiveBayesModel("naivebayes-uid", pi, theta)
+    // TODO(@sid.murching/reviewers) This functionality could be
+    // provided in ModelBuilder.scala; however, it seems we need to set the modelType param
+    // (a private[classification] member) explicitly. Otherwise, when the NaiveBayesModel
+    // constructed here is used, the default parameter setting for modelType
+    // is not detected & an exception is thrown.
+    model.set(model.modelType, "multinomial")
```
smurching: Weird -- I'm not sure why I was running into access privilege issues before, but I was able to just set the `modelType` param in `ModelBuilder.newNaiveBayesModel`, so I did that instead of adding a `setParam` interface to `TrainingSetFromTransformer`.

src/main/scala/com/databricks/spark/sql/perf/mllib/MLTransformerBenchmarkable.scala
```diff
@@ -48,17 +49,27 @@ class MLTransformerBenchmarkable(
...
-        val estimator = test.getEstimator(param)
-        estimator.fit(trainingData)
+        test.getPipelineStage(param) match {
+          case est: Estimator[_] => est.fit(trainingData)
+          case transformer: Transformer =>
+            transformer.transform(trainingData)
+            transformer
+          case other: Any => throw new UnsupportedOperationException("Algorithm to benchmark must" +
+            s" be an estimator or transformer, found ${other.getClass} instead.")
+        }
```
smurching: Done.

src/main/scala/com/databricks/spark/sql/perf/mllib/feature/Bucketizer.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.spark.sql._
+
+import com.databricks.spark.sql.perf.mllib.OptionImplicits._
+import com.databricks.spark.sql.perf.mllib.{BenchmarkAlgorithm, MLBenchContext, TestFromTraining}
+
+/** Object for testing Bucketizer performance */
+object Bucketizer extends BenchmarkAlgorithm with TestFromTraining with UnaryTransformer {
+
+  // Min/max possible Double values for input column to Bucketizer
+  private val (maxBucketizerVal, minBucketizerVal) = (-1e9, 1e9)
```
jkbradley: min/max are switched

src/main/scala/com/databricks/spark/sql/perf/mllib/feature/Bucketizer.scala
```diff
@@ -0,0 +1,47 @@
...
+  private def doubleInRange(rng: Random, lo: Double, hi: Double): Double = {
+    rng.nextDouble() * (hi - lo) + lo
+  }
+
+  override def trainingDataSet(ctx: MLBenchContext): DataFrame = {
+    import ctx.params._
+    import ctx.sqlContext.implicits._
+    val rng = ctx.newGenerator()
+    // For a bucketizer, training data consists of a single column of random doubles
+    val colValsToBucketize = 0L.until(numExamples).map(_ => doubleInRange(rng,
```
jkbradley: This is not a very scalable way to generate the data: it's generating it on the driver and then distributing it.  I'd recommend reusing DataGenerator.generateContinuousFeatures.  If we choose reasonable split points in [-1,+1], then it should work well.

src/main/scala/com/databricks/spark/sql/perf/mllib/feature/Bucketizer.scala
```diff
@@ -0,0 +1,47 @@
...
+    import ctx.params._
+    val rng = ctx.newGenerator()
+    // Generate an array of (finite) splitting points for the Bucketizer
+    val splitPoints = 0.until(bucketizerNumBuckets - 1).map { _ =>
+      doubleInRange(rng, minBucketizerVal, maxBucketizerVal)
+    }.sorted.toArray
+    // Final array of splits contains +/- infinity
+    val splits = Array(Double.NegativeInfinity) ++ splitPoints ++ Array(Double.PositiveInfinity)
+    new ml.feature.Bucketizer()
+        .setSplits(splits)
```
jkbradley: style: indent -2 spaces

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/NaiveBayes.scala
```diff
@@ -0,0 +1,65 @@
...
+  with TestFromTraining with TrainingSetFromTransformer with ScoringWithEvaluator {
+
+  override protected def initialData(ctx: MLBenchContext) = {
+    import ctx.params._
+    val rng = ctx.newGenerator()
+    // Max possible arity of a feature in generated training/test data for NaiveBayes models
+    val maxFeatureArity = 20
+    // All features for Naive Bayes must be categorical, i.e. have arity >= 2
+    val featureArity = 0.until(numFeatures).map(_ => 2 + rng.nextInt(maxFeatureArity - 2)).toArray
+    val df = DataGenerator.generateMixedFeatures(
```
jkbradley: style: no need to define temp val "df"

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/NaiveBayes.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  override protected def trueModel(ctx: MLBenchContext): Transformer = {
+    import ctx.params._
+    val rng = ctx.newGenerator()
+    // pi = log of class priors, whose dimension is C (number of classes)
+    // theta = log of class conditional probabilities, whose dimension is C (number of classes)
+    // by D (number of features)
+    val unnormalizedProbs = 0.until(numClasses).map(_ => rng.nextDouble() + 1e-5).toArray
+    val piArray = unnormalizedProbs.map(prob => math.log(prob / unnormalizedProbs.sum))
```
jkbradley: It does not matter here, but still, I'd move the sum outside of the map to make sure it's only computed once.

jkbradley: You can also write it as log(prob) - log(sum) to make it more numerically stable.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/NaiveBayes.scala
```diff
@@ -0,0 +1,65 @@
...
+    val thetaArray = Array.tabulate(numClasses) { i: Int =>
+      val baseProbMass = (1 - currClassProb) / (numFeatures - 1)
+      val probs = Array.fill[Double](numFeatures)(baseProbMass)
+      probs(i) = currClassProb
+      probs
+    }.map(_.map(math.log))
+
+    // Initialize new Naive Bayes model
+    val pi = Vectors.dense(piArray)
+    val theta = new DenseMatrix(numClasses, numFeatures, thetaArray.flatten, true)
```
jkbradley: Can you please double-check to make sure this is generated correctly?  I think there is a mistake: The DenseMatrix constructor should have numClasses and numFeatures switched...I think, and then you'd take the transpose to get theta.

smurching: Thanks for checking, I think it's correct as is - the matrix should be `numClasses` by `numFeatures`, where each row corresponds to the (log) conditional distribution of the features for a given class.

Setting the final constructor arg (`isTranspose`) to `true` indicates that the matrix's values are in row-major order, which afaik is how thetaArray.flatten() works.

NaiveBayes.train() sets theta similarly ([link](https://github.com/apache/spark/blob/60022bfd65e4637efc0eb5f4cc0112289c783147/mllib/src/main/scala/org/apache/spark/ml/classification/NaiveBayes.scala#L202))

src/main/scala/com/databricks/spark/sql/perf/mllib/MLLib.scala
```diff
@@ -46,13 +53,16 @@ object MLLib extends Logging {
       require(yamlConfig != null)
       YamlConfig.readString(yamlConfig)
     }
+
+    // Uncomment these lines to run locally:
+    // val sparkConf = new SparkConf().setAppName("MLlib QA").setMaster("local[2]")
+    // val sc = new SparkContext(sparkConf)
     val sc = SparkContext.getOrCreate()
-    sc.setLogLevel("INFO")
```
smurching: No, not sure why I deleted it; I've added it back in

smurching:
Oops, was testing using my own YAML file; I've added its contents to mllib-small.yaml.

src/main/scala/com/databricks/spark/sql/perf/mllib/classification/NaiveBayes.scala
jkbradley: OH, you're right, I missed the isTransposed

jkbradley:
LGTM, merging, thanks!

jkbradley:
@marmbrus @yhuai @JoshRosen Could you please sign off on the changes which affect the SQL tests?  They are all about running with Spark 2.2:
* build.sbt
* src/main/scala/com/databricks/spark/sql/perf/Benchmark.scala
* src/main/scala/com/databricks/spark/sql/perf/Query.scala

yhuai:
@juliuszsompolski @bogdanrdc With this change, spark-sql-perf will not compile with spark 2.1. I think it is fine but just want to double check with you guys.

juliuszsompolski:
@yhuai By "don't compile" you mean just compilation, or it won't run e.g. TPCDS on 2.1 clusters?
We currently still use runs against 2.1.1-db4 as baseline, but I guess that could be moved forward. @harsha2010 

If we are to break compatibility with 2.1, maybe let's branch the current state as branch-2.1, and update the readme to:
"This is a performance testing framework for Spark SQL in Apache Spark 2.2+."
and add something like:
"For support of earlier versions of Apache Spark, see branches in the repository."

juliuszsompolski:
I see there is already a branch-0.2 for a version supporting Spark 1.4+, and the current version.sbt is at 0.4.12, so maybe
* cut it as branch-0.4 (removing -SNAPSHOT from version.sbt there)
* bump version to 0.5.0-SNAPSHOT in master.

yhuai:
Yea, sounds good. I guess we can keep the `-SNAPSHOT` part in the version.sbt in branch-0.4 in case we still need to merge anything to that branch.

@smurching Can we do what @juliuszsompolski is suggesting? Specifically, let do:
- Cut a branch-0.4 from the current master branch.
- In the master branch, we change version number in version.sbt to -.5.0-SNAPSHOT.
- In this pr, mention that it requires spark 2.2+.

Then, we can get this pr merged. Thanks!

yhuai:
lgtm

jkbradley:
Merging!

PR #112, opened 2017-08-22
Add benchmark for LinearSVC/OnehotEncoder/VectorSlicer/VectorAssembler/StringIndexer/Tokenizer

WeichenXu123:
Add benchmark for: 
 - LinearSVC
 - OnehotEncoder
 - VectorSlicer
 - VectorAssembler
 - StringIndexer
 - Tokenizer

FPGrowth is working on will be added soon.

Part of code from https://github.com/smurching/spark-sql-perf/pull/1 and address feedbacks from @smurching

PR #114, opened 2017-08-25
Refactor MLParams for spark-sql-perf

smurching:
A case class (MLParams) is currently used to store/access parameters for ML tests in spark-sql-perf. With the addition of new ML tests to spark-sql-perf (in this PR: https://github.com/databricks/spark-sql-perf/pull/112), the number of ML-related test params will be > 22, but Scala only allows up to 22 params in a case class.

This PR addresses the issue by:
* Introducing a new MLParameters class (`class MLParameters`) that provides access to the same parameters as MLParams, except as a class instead of a case class.
* Replacing usages of MLParams with MLParameters
* Storing the members of MLParameters in BenchmarkResult.parameters for logging/persistence.

Tested by running default performance tests in `src/main/scala/configs/mllib-small.yaml`.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
@@ -1,10 +1,14 @@
 package com.databricks.spark.sql.perf.mllib
 
-import scala.reflect.ClassTag
+import scala.reflect.{classTag, ClassTag}
 import scala.reflect.runtime.universe._
 
-/** Exposes methods to simplify implementation of classes like MLParameters. */
+/** Exposes methods to simplify implementation of classes like MLParams. */
 private[perf] object ReflectionUtils {
+
+  private def getConstructor[T: TypeTag: ClassTag](obj: T): MethodSymbol = {
+    typeOf[T].declaration(nme.CONSTRUCTOR).asMethod
+  }
 
   /**
    * Given an instance [[obj]] of a class whose constructor arguments are all of type Option[Any],
@@ -14,7 +18,7 @@ private[perf] object ReflectionUtils {
    */
   def getConstructorArgs[T: TypeTag: ClassTag](obj: T): Map[String, Any] = {
     // Get constructor of passed-in instance
-    val constructor = typeOf[T].declaration(nme.CONSTRUCTOR).asMethod
+    val constructor = getConstructor(obj)
     // Include each constructor argument not equal to None in the output map
     constructor.paramss.flatten.flatMap { (param: Symbol) =>
       // Get name and value of the constructor argument
@@ -30,6 +34,21 @@ private[perf] object ReflectionUtils {
           s"type Option[Any]; constructor argument ${paramName} had invalid type.")
       }
     }.toMap
+  }
+
+  /**
+   * Create a copy of [[obj]] by calling obj's constructor with the same parameters as
+   * those used to instantiate [[obj]]
+   */
+  def copy[T: TypeTag: ClassTag](obj: T): T = {
+    // Get arguments passed to obj's constructor
+    val constructorArgValues = getConstructorArgs(obj).values.toSeq
+    // Create runtime mirror
+    val rm = runtimeMirror(classTag[T].runtimeClass.getClassLoader)
+    val classMirror = rm.reflectClass(typeOf[T].typeSymbol.asClass)
+    // Create new instance with same class as obj
+    val constructor = getConstructor(obj)
+    classMirror.reflectConstructor(constructor).apply(constructorArgValues: _*).asInstanceOf[T]
   }
 
 }
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
@@ -14,9 +14,9 @@ private[perf] object ReflectionUtils {
    */
   def getConstructorArgs[T: TypeTag: ClassTag](obj: T): Map[String, Any] = {
     // Get constructor of passed-in instance
-    val constructor = typeOf[T].decl(termNames.CONSTRUCTOR).asMethod
+    val constructor = typeOf[T].declaration(nme.CONSTRUCTOR).asMethod
     // Include each constructor argument not equal to None in the output map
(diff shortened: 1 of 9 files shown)


src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
```diff
@@ -0,0 +1,35 @@
...
+/** Exposes methods to simplify implementation of classes like MLParameters. */
+private[perf] object ReflectionUtils {
+
+  /**
+   * Given an instance [[obj]] of a class whose constructor arguments are all of type Option[Any],
+   * returns a map of key-value pairs (argName -> argValue) where argName is the name
+   * of a constructor argument with a defined (not None) value and argValue is the corresponding
+   * value.
+   */
+  def getConstructorArgs[T: TypeTag: ClassTag](obj: T): Map[String, Any] = {
```
smurching: This method allows me to avoid a bunch of statements of the form
```
if (parameter.isDefined) {
  map ++= (parameter -> numFeatures.get.toString)
}
```
in `MLParameters.toMap` -- let me know if it'd be better to not use reflection/do this another way.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -104,8 +106,93 @@ case class BreakdownResult(
...
+
+/**
+ * DEPRECATED IN FAVOR OF MLParameters.
+ *
+ * Case class wrapping parameters for ML tests.
+ *
+ * KEEP CONSTRUCTOR ARGUMENTS SORTED BY NAME.
+ * It simplifies lookup when checking if a parameter is here already.
+ */
 case class MLParams(
```
WeichenXu123: So why don't mark this class `@deprecate` ?

src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
WeichenXu123: Hmm... interesting, use reflection here, can we also use reflection to generate the `MLParameters.copy` method ? So we can avoid duplicate code.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -104,8 +106,93 @@ case class BreakdownResult(
...
+      numTestExamples: Option[Long] = numTestExamples,
+      numPartitions: Option[Int] = numPartitions,
+      // *** Specialized and sorted by name ***
+      bucketizerNumBuckets: Option[Int] = bucketizerNumBuckets,
+      depth: Option[Int] = depth,
+      elasticNetParam: Option[Double] = elasticNetParam,
+      family: Option[String] = family,
+      k: Option[Int] = k,
+      ldaDocLength: Option[Int] = ldaDocLength,
+      ldaNumVocabulary: Option[Int] = ldaNumVocabulary,
```
WeichenXu123: rename to `docLength` and `numVocabulary` ? cc @jkbradley

src/main/scala/com/databricks/spark/sql/perf/mllib/ReflectionUtils.scala
smurching: I tried to do this but it got messy so I just stuck with writing out the copy constructor :P

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -104,37 +105,87 @@ case class BreakdownResult(
 
 case class Failure(className: String, message: String)
 
-// KEEP ARGUMENTS SORTED BY NAME.
-// It simplifies lookup when checking if a parameter is here already.
-case class MLParams(
+/**
+ * Class wrapping parameters for ML tests; provides the same functionality as MLParams, but is
```
jkbradley: Update this text: You are talking about the old MLParams, which won't mean anything to people reading the docs.  That kind of comment belongs in PR comments, not in the code.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -104,37 +105,87 @@ case class BreakdownResult(
...
-)
+    val bucketizerNumBuckets: Option[Int] = None,
+    val depth: Option[Int] = None,
+    val docLength: Option[Int] = None,
+    val elasticNetParam: Option[Double] = None,
+    val family: Option[String] = None,
+    val k: Option[Int] = None,
+    val link: Option[String] = None,
+    val maxIter: Option[Int] = None,
+    val naiveBayesSmoothing: Option[Double] = None,
```
jkbradley: As long as you're fixing names, can you rename this to "smoothing" since I bet we'll have other algs use it in the future.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -104,37 +105,87 @@ case class BreakdownResult(
...
+    val numClasses: Option[Int] = None,
+    val numFeatures: Option[Int] = None,
+    val numItems: Option[Int] = None,
+    val numUsers: Option[Int] = None,
+    val numVocabulary: Option[Int] = None,
```
jkbradley: rename to "vocabSize" to match MLlib

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -104,37 +105,87 @@ case class BreakdownResult(
...
+      numClasses: Option[Int] = numClasses,
+      numFeatures: Option[Int] = numFeatures,
+      numItems: Option[Int] = numItems,
+      numUsers: Option[Int] = numUsers,
+      numVocabulary: Option[Int] = numVocabulary,
+      optimizer: Option[String] = optimizer,
+      regParam: Option[Double] = regParam,
+      rank: Option[Int] = rank,
+      tol: Option[Double] = tol): MLParams = {
+    new MLParams(randomSeed, numExamples, numTestExamples, numPartitions, bucketizerNumBuckets,
```
jkbradley: This will be easy to mess up later, so how about passing args by name instead of order?

smurching:
I experimented with adding ~20 more arguments to MLParams, the tests still compile and run.

jkbradley:
LGTM
Merging with master
Thanks @smurching !

WeichenXu123:
Great! I will update my PRs soon!

PR #116, opened 2017-08-31
[ML-2847] Add new tests for (DecisionTree, RandomForest)Regression, GMM, HashingTF

smurching:
This PR follows up on #112, adding new performance tests for DecisionTreeRegression, RandomForestRegression, GMM, and HashingTF.

Summary of changes:
* Added new performance tests
* Updated configs in mllib-small.yaml
    * Alphabetized configs
    * Added new configs for: RandomForestRegression, DecisionTreeRegression, GMM, HashingTF
* Refactored `TreeOrForestClassification` into a trait (`TreeOrForestEstimator`) exposing methods for all tree/forest estimator performance tests. 
    * Copied code from `DecisionTreeClassification.scala` into `TreeOrForestEstimator.scala`

I tested this PR by running the performance tests specified in `mllib-small.yaml`


smurching:
This PR looks kinda big since I alphabetized the configs in mllib-small.yaml (to help keep track of which tests were being run). Let me know if it'd be helpful for me to de-alphabetize them/just commit the new test configs I've added.

src/main/scala/configs/mllib-small.yaml
```diff
@@ -112,10 +106,44 @@ benchmarks:
...
+      tol: 0.0
+      maxIter: 10
+      regParam: 0.1
+  - name: regression.LinearRegression
+    params:
+      numExamples: 100
+      numTestExamples: 100
+      numFeatures: 100
+      regParam: 0.1
+      tol: [0.0]
```
WeichenXu123: The `tol` seems should > 0 ?

src/main/scala/com/databricks/spark/sql/perf/mllib/feature/HashingTF.scala
```diff
@@ -0,0 +1,41 @@
...
+    // each string is selected from a pool of vocabSize strings
+    // The expected # of occurrences of each word in our vocabulary is
+    // (docLength * numExamples) / vocabSize
+    DataGenerator.generateDoc(ctx.sqlContext, numExamples = numExamples, seed = ctx.seed(),
+      numPartitions = numPartitions, vocabSize = vocabSize, avgDocLength = docLength, inputCol)
+  }
+
+  override def getPipelineStage(ctx: MLBenchContext): PipelineStage = {
+    import ctx.params._
+    val rng = ctx.newGenerator()
```
WeichenXu123: `rng` not used. Can be removed.

src/main/scala/com/databricks/spark/sql/perf/mllib/feature/HashingTF.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  override def trainingDataSet(ctx: MLBenchContext): DataFrame = {
+    import ctx.params._
+    // To test HashingTF, we generate arrays of (on average) docLength strings, where
+    // each string is selected from a pool of vocabSize strings
+    // The expected # of occurrences of each word in our vocabulary is
+    // (docLength * numExamples) / vocabSize
+    DataGenerator.generateDoc(ctx.sqlContext, numExamples = numExamples, seed = ctx.seed(),
+      numPartitions = numPartitions, vocabSize = vocabSize, avgDocLength = docLength, inputCol)
```
WeichenXu123: `inputCol` ==> `dataColName = inputCol`
I think the normal parameters should be put before `name = value` style parameters. Otherwise having risk of being ambiguous.

src/main/scala/com/databricks/spark/sql/perf/mllib/TreeOrForestEstimator.scala
```diff
@@ -0,0 +1,74 @@
...
+}
+
+private[mllib] object TreeOrForestEstimator {
+
+  /**
+   * Get feature arity for tree and tree ensemble tests.
+   * Currently, this is hard-coded as:
+   * - 1/2 binary features
+   * - 1/2 high-arity (20-category) features
+   * - 1/2 continuous features
```
WeichenXu123: should be `1/4`, `1/4`, `1/2`

src/main/scala/com/databricks/spark/sql/perf/mllib/clustering/GaussianMixture.scala
```diff
@@ -0,0 +1,29 @@
+package com.databricks.spark.sql.perf.mllib.clustering
+
+import org.apache.spark.ml
+import org.apache.spark.ml.PipelineStage
+import org.apache.spark.sql.DataFrame
+
+import com.databricks.spark.sql.perf.mllib.OptionImplicits._
```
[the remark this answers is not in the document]
smurching: Hm what's the import ordering issue here? `com.databricks.spark.sql.perf.mllib.OptionImplicits` does have "spark" in its classname but it is in the `com.databricks` package, so it should follow the Spark imports.

src/main/scala/configs/mllib-small.yaml
smurching: Good question -- I think if `tol` is 0, the algorithm will just run for `maxIter` iterations (LBFGS, WeightedLeastSquares, OWLQN optimizers all take `maxIter` as a constructor argument).

smurching: I don't think it's super important what `tol` is set to here (since this config just contains small parameters to verify all of the performance tests run properly), but I'll update it to 1e-6 (the Spark default for LinearRegression).

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -124,6 +124,7 @@ class MLParams(
     val elasticNetParam: Option[Double] = None,
     val family: Option[String] = None,
     val featureArity: Option[Int] = None,
+    val featurizerOutputDim: Option[Int] = None, // # of output features produced by a featurizer
```
jkbradley: There's no need to add a new parameter.  Just reuse numFeatures

src/main/scala/com/databricks/spark/sql/perf/mllib/clustering/GaussianMixture.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.databricks.spark.sql.perf.mllib.{BenchmarkAlgorithm, MLBenchContext, TestFromTraining}
+import com.databricks.spark.sql.perf.mllib.data.DataGenerator
+
+object GaussianMixture extends BenchmarkAlgorithm with TestFromTraining {
+
+  override def trainingDataSet(ctx: MLBenchContext): DataFrame = {
+    import ctx.params._
+    DataGenerator.generateContinuousFeatures(ctx.sqlContext, numExamples, ctx.seed(),
```
jkbradley: We have a DataGenerator.generateGaussianMixtureData method, so I'd use that

src/main/scala/com/databricks/spark/sql/perf/mllib/feature/HashingTF.scala
```diff
@@ -0,0 +1,41 @@
...
+    Array.tabulate[String](length)(_ => dictionary(rng.nextInt(dictLength)))
+  }
+
+  override def trainingDataSet(ctx: MLBenchContext): DataFrame = {
+    import ctx.params._
+    // To test HashingTF, we generate arrays of (on average) docLength strings, where
+    // each string is selected from a pool of vocabSize strings
+    // The expected # of occurrences of each word in our vocabulary is
+    // (docLength * numExamples) / vocabSize
+    DataGenerator.generateDoc(ctx.sqlContext, numExamples = numExamples, seed = ctx.seed(),
```
jkbradley: This will generate each row as a single string, but you want to break rows into ```Array[String]```.  Did you mean to use the randomSentence method?  I'd recommend use generateDoc as you are and then calling something like this to split the doc:
```
import org.apache.spark.sql.functions.split
df.withColumn(inputCol, split(inpuCol, " "))
```

jkbradley:
LGTM
I ran it locally
Merging with master
Thanks @smurching and @WeichenXu123 !

PR #117, opened 2017-09-01
fix `df.drop` in `VectorAssembler`

WeichenXu123:
fix `df.drop` in `VectorAssembler` in #112


WeichenXu123:
cc @jkbradley @smurching

PR #109, opened 2017-07-14
Support for TPC-H benchmark

bogdanrdc:
* Refactoring of TPC-DS code to be able to reuse it for TPC-H.
* TPC-H queries texts adapted for Spark

- [x] this requires support for dbgen stdout patch, need to publish that somewhere stable
- [x] check if breaks existing scripts

PR #118, opened 2017-09-04
Coalesce non-partitioned tables.

juliuszsompolski:
In https://github.com/databricks/spark-sql-perf/pull/109/ coalescing of non-partitioned tables into 1 file seems to have gotten accidentally removed.
Put it back, but only when clusterByPartitionedColumns == true
Considering that we coalesce partitions only when that setting is true, it seems to be consistent to use it also for non-partitioned tables.

It may be better to change the name of the parameter, but that changes the interface, and possibly should be left for some future clean up.


bogdanrdc:
lgtm

PR #119, opened 2017-09-12
Don't clean blocks after every run in generic Benchmarkable

juliuszsompolski:
Do not clean blocks after each run in the generic Benchmarkable trait.
It seems to have been there since #33, and an option `spark.databricks.benchmark.cleanBlocksAfter` to turn it off was added to it in #98, specifically to allow parallel TPCDS streams to not wipe each other's blocks. But that option is quite well hidden and obscure, and as a SparkContext config option can only be set during cluster creation, so it's not friendly to use.

Cleaning up the blocks doesn't seem necessary for the Query Benchmarkables used for TPCDS and TPCH. Remove it from there, and leave it only for MLPipelineStageBenchmarkable.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.5.0-SNAPSHOT"
+version in ThisBuild := "0.5.1"
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.5.1"
+version in ThisBuild := "0.5.0-SNAPSHOT"
--- a/src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Benchmarkable.scala
@@ -55,15 +55,8 @@ trait Benchmarkable extends Logging {
 
   protected def beforeBenchmark(): Unit = { }
 
-  private def afterBenchmark(sc: SparkContext): Unit = {
-    // Best-effort clean up of weakly referenced RDDs, shuffles, and broadcasts
+  protected def afterBenchmark(sc: SparkContext): Unit = {
     System.gc()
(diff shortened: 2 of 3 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.5.0-SNAPSHOT"
+version in ThisBuild := "0.5.1"
```
juliuszsompolski: we should start just bumping version with every code change, because later it gets hard to keep track of what exact version the given jar is...

juliuszsompolski:
cc @jkbradley You wrote me that cleaning of the blocks was important for ML workloads. Is this change ok for these workloads?
cc @ericl and @JoshRosen You merged in the `spark.databricks.benchmark.cleanBlocksAfter`. Are you aware of any workloads this change may break?

ericl:
Sounds good

version.sbt
jkbradley: That seems pretty error-prone.  Why not always use versions of spark-sql-perf produced as actual releases (with corresponding git tags)?  Is publishing too heavy weight a process?

jkbradley:
Sorry for the slow response!  This change seems fine to me, except for the comment above about changing version numbers.

version.sbt
juliuszsompolski: Over the last few months we kept merging various small changes, including ones that changed some interfaces (all around TPCDS benchmark) without changing the version number. So now there are various jars circulating that are hard to distinguish what code version they are precisely. Most of this mess is my fault, apologies.
I figured out that maybe bumping the minor version with every code change made is a good idea, but I have no idea what are the actual best practices of doing that re: git tags / branches, version dependencies for maven etc. etc.
So for this change, since it's unrelated, let's just keep the version and push it that is.

Going forward @npoggi will be managing spark-sql-perf so I'll leave to him figuring out what is the best practice way to version and release it :-).

PR #122, opened 2017-09-20
run full gc on the execturos after each run

liufengdb:
I did not find a better to make sure gc command is running on all the executors.


juliuszsompolski:
For a committed solution I'd make it an optional parameter. Otherwise it may interfere with e.g. parallel runs. I hope that running a custom jar with this hacked in unblocks you for right now.

PR #123, opened 2017-11-07
Update README to do sql("use database")

juliuszsompolski:


juliuszsompolski:
I wouldn't bother in the notebooks - sql( and %sql are the same.

PR #125, opened 2018-02-21
[ML-3342] Bug fixes to make mllib benchmarks work with dbr-4.0.

MrBago:
In spark 2.3 some default param values were moved from Models to matching Estimators. I added explicit sets for these values in our tests to avoid errors. Also renamed ModelBuilder to ModelBuildersSSP to avoid a name conflict with dbml-local which is included in databricks runtime.

### DIFF
--- a/src/main/scala/org/apache/spark/ml/ModelBuilderSSP.scala
+++ b/src/main/scala/org/apache/spark/ml/ModelBuilderSSP.scala
@@ -17,6 +17,7 @@ object ModelBuilderSSP {
       coefficients: Vector,
       intercept: Double): LogisticRegressionModel = {
     new LogisticRegressionModel("lr", coefficients, intercept)
+      .setThreshold(.5)
 
   }
 
--- a/src/main/scala/org/apache/spark/ml/ModelBuilderSSP.scala
+++ b/src/main/scala/org/apache/spark/ml/ModelBuilderSSP.scala
@@ -11,12 +11,13 @@ import org.apache.spark.mllib.tree.impurity.ImpurityCalculator
 /**
  * Helper for creating MLlib models which have private constructors.
  */
-object ModelBuilder {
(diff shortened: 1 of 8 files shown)


src/main/scala/org/apache/spark/ml/ModelBuilderSSP.scala
```diff
@@ -11,18 +11,24 @@ import org.apache.spark.mllib.tree.impurity.ImpurityCalculator
...
+      .setThreshold(.5)
+
   }
 
   def newLinearRegressionModel(
       coefficients: Vector,
       intercept: Double): LinearRegressionModel = {
-    new LinearRegressionModel("linr", coefficients, intercept)
+    val model = new LinearRegressionModel("linr", coefficients, intercept)
+    if (model.hasParam("loss")) {
```
smurching: Could we remove this `if` by updating `build.sbt` to build spark-sql-perf against Spark 2.3 instead of 2.2?

juliuszsompolski:
cc @npoggi

src/main/scala/org/apache/spark/ml/ModelBuilderSSP.scala
npoggi: @smurching Spark 2.3 is not out yet.  I would keep the `if` at least half a year more.  The TPC-* part of `spark-sql-perf` works fine built against 2.2.

smurching: Gotcha, yeah that sounds good to me - thanks for clarifying!

smurching:
Given the discussion above I think this PR is good as is, thanks @MrBago!

jkbradley:
LGTM too.  @juliuszsompolski or @npoggi do you want to vet all spark-sql-perf PRs, or do you mind if I merge ones for ML?

PR #128, opened 2018-02-26
MinHashLSH and BucketedRandomProjectionLSH benchmark

WeichenXu123:
### What's the PR

MinHashLSH and BucketedRandomProjectionLSH benchmark added.

### Discussion

- Whether we need to improve the way of testing data generation for MinHashLSH ( and add more control param, such as max/min element number in each input set )

- Whether we need to add benchmark for `approxNearestNeighbors` and `approxSimilarityJoin`


MrBago:
LGTM! Thanks @WeichenXu123

MrBago:
I was able to run this locally, I think we can merge this :).

jkbradley:
Thanks all!  I'll merge it.

PR #129, opened 2018-03-07
Remove VACUUM from tpcds_datagen notebook.

juliuszsompolski:
VACUUM is not needed after data generation.


juliuszsompolski:
cc @npoggi

npoggi:
@juliuszsompolski thanks, merge it as you can.

PR #130, opened 2018-03-07
Use DECIMAL and DATE in the default TPCDS notebooks.

juliuszsompolski:
A long time ago, DECIMAL was substituted by DOUBLE and DATE by STRING to workaround some problems.
There is no reason to do it anymore.


juliuszsompolski:
cc @npoggi

npoggi:
@juliuszsompolski perfect, we need this so the docs and data generation are more straightforward.  Will update my derived notebooks to use the new DDL.

PR #137, opened 2018-03-22
Adds an optional version of the SS_MAX query

npoggi:
Adds ss_maxb query removing the `count(distinct)` to highlight more I/O speed.
It only runs if manually selected.


src/main/resources/tpcds_2_4/ss_max_b.sql
```diff
@@ -0,0 +1,14 @@
+select
```
juliuszsompolski: You seem to still have two copies of the file committed after the rename.

PR #138, opened 2018-03-22
Coalesce(n) instead of hardcoded (1) for large tables/partitions

npoggi:
When using `clusterByPartitionColumns` to coalesce into multiple files (instead of hardcoded `1`) when the number of records is larger than `spark.sql.files.maxRecordsPerFile`.  It has the cost of a count operation, but enables multiple writers for large scale factors.  This improves cluster utilization and reduces total generation time, in especial to non-partitioned tables (or with few partitions i.e., TPCH) and large scale factors.  The cost of count is leveraged as partitioning the data is needed anyway for the multiple writers, skipping a stage.  This feature is only enabled if  `spark.sql.files.maxRecordsPerFile` is used,  not by default in spark, but hardcoded in the datagen notebooks to produce large files and deterministic structure (repeatable using different cluster sizes).

Additionaly updates deprecated `registerTempTable` to `createOrReplaceTempView` to avoid warnings.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/Tables.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Tables.scala
@@ -179,7 +179,7 @@ abstract class Tables(sqlContext: SQLContext, scaleFactor: String,
 
       val data = df(format != "text", numPartitions)
       val tempTableName = s"${name}_text"
-      data.registerTempTable(tempTableName)
+      data.createOrReplaceTempView(tempTableName)
 
       val writer = if (partitionColumns.nonEmpty) {
         if (clusterByPartitionColumns) {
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/spark/sql/perf/Tables.scala
```diff
@@ -211,9 +211,24 @@ abstract class Tables(sqlContext: SQLContext, scaleFactor: String,
           data.write
         }
       } else {
+        // treat non-partitioned tables as "one partition" that we want to coalesce
         if (clusterByPartitionColumns) {
-          // treat non-partitioned tables as "one partition" that we want to coalesce
-          data.coalesce(1).write
+          // in case data has more than maxRecordsPerFile, split into multiple writers to improve datagen speed
+          // files will be truncated to maxRecordsPerFile value, so the final result will be the same
+          val numRows = data.count
```
juliuszsompolski: Hmm... It's beneficial, but a bit ugly that this .count will cause the dsdgen / tpch datagen to be executed and do all the generating job twice.
Maybe for known scale factors you could hardcode a lookup table (tablename, sf) -> count, empty in Tables.scala, and overriden in TPCDS.scala and TPCH.scala?
Dirty few lines of code that could further shave a lot of time on high SF.

npoggi:
@juliuszsompolski I finally had time to benchmark the benefit of this PR.  On a TPCDS SF1,000 the time to generate and write the data to parquet is roughly the same with count + coalesce(n) vs coalesce(1):  2.48hrs vs 2.55.  This is due to TPCDS being highly partitioned already, as this PR only changes the behavior for non-partitioned tables.  On the extreme case of no-partitioning at SF1,000; the time would be 3.7hrs vs more than 21hrs (and still running).
For TPCH which was why I had tested this approach, at SF1,000 with count + coalesce(n) vs coalesce(1): 2.04hrs vs 2.99.  I was higher even at SF10,000 where the jobs were not finishing on the cluster size I had before this change.
In summary, there is no loss in performance for TCPDS and better cluster utilization and time improvements.  Will go ahead to merge it.

PR #140, opened 2018-05-08
Add decision tree benchmark

MrBago:
This PR moves the mllib-config file to resources and adds a large config file for running larger production benchmarks.


src/main/scala/com/databricks/spark/sql/perf/mllib/MLLib.scala
```diff
@@ -34,6 +36,21 @@ object MLLib extends Logging {
     e.getCurrentResults()
   }
 
+  private def getConfig(resourcePath: String): String = {
+    val stream = getClass.getResourceAsStream(resourcePath)
+    val buffer = new Array[Byte](10000)
```
mengxr: try `scala.io.Source.fromInputStream(...)`

src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
```diff
@@ -0,0 +1,13 @@
+output: /databricks/spark/sql/mllib-perf-ci
+timeoutSeconds: 1000
```
mengxr: just to confirm, is this timeout good for the large test?

MrBago: It's enough to benchmark 1 estimator, but we'll need to bump it up as we add more.  I can add a comment.

MrBago:
@mengxr can you take another look.

PR #143, opened 2018-05-23
Added benchmarks to mllib-large.yaml for classifcation Estimators.

MrBago:


mengxr:
@MrBago Could you comment on where you get the param values? Are they copied from previous test suites or new?

PR #144, opened 2018-06-01
[ML-3749] Log metric name and isLargerBetter in BenchmarkResult

ludatabricks:
- Add new case class "MLMetrics" to save all different metrics

- Change "mlResult" in BenchmarkResult to Array[MLMetrics] 

- score function will return MLMetrics

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
@@ -7,6 +7,8 @@ import org.apache.spark.ml.evaluation.Evaluator
 import org.apache.spark.sql._
 import org.apache.spark.sql.functions._
 
+import com.databricks.spark.sql.perf._
+
 /**
  * The description of a benchmark for an ML algorithm. It follows a simple, standard proceduce:
  *  - generate some test and training data
@@ -38,7 +40,7 @@ trait BenchmarkAlgorithm extends Logging {
   def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): Double = -1.0 // Not putting NaN because it is not valid JSON.
+      model: Transformer): MLMetrics = new MLMetrics()
 
   def name: String = {
     this.getClass.getCanonicalName.replace("$", "")
@@ -67,9 +69,10 @@ trait ScoringWithEvaluator {
   final override def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): Double = {
+      model: Transformer): MLMetrics = {
     val eval = model.transform(testSet)
-    evaluator(ctx).evaluate(eval)
+    // TODO: add MetricName for evaluators
+    new MLMetrics("", evaluator(ctx).evaluate(eval), evaluator(ctx).isLargerBetter)
   }
 }
 
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
@@ -40,7 +40,7 @@ trait BenchmarkAlgorithm extends Logging {
   def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): MLMetrics = new MLMetrics()
+      model: Transformer): MLMetric = MLMetric.Invalid
 
   def name: String = {
     this.getClass.getCanonicalName.replace("$", "")
@@ -69,10 +69,17 @@ trait ScoringWithEvaluator {
   final override def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): MLMetrics = {
-    val eval = model.transform(testSet)
-    // TODO: add MetricName for evaluators
-    new MLMetrics("", evaluator(ctx).evaluate(eval), evaluator(ctx).isLargerBetter)
+      model: Transformer): MLMetric = {
+    val results = model.transform(testSet)
+    val eval = evaluator(ctx)
+    val metricName = if (eval.hasParam("metricName")) {
+      val param = eval.getParam("metricName")
+      eval.getOrDefault(param).toString
+    } else {
+      eval.getClass.getSimpleName
+    }
+    val metricValue = eval.evaluate(results)
+    MLMetric(metricName, metricValue, eval.isLargerBetter)
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -223,19 +223,13 @@ object MLParams {
...
+ * @param metricValue the value of the metric
+ * @param isLargerBetter  (MLlib) Test metric, such as accuracy
  */
-case class MLResult(
-    trainingTime: Option[Double] = None,
-    trainingMetric: Option[Double] = None,
-    testTime: Option[Double] = None,
-    testMetric: Option[Double] = None,
-    additionalTests: Map[String, Double])
+case class MLMetrics(
```
mengxr: This is a single metric. We should call it `MLMetric`.

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -223,19 +223,13 @@ object MLParams {
...
+    metricName: String = null,
```
mengxr: We should remove the default values and let user specify them explicitly. We can provide a "Invalid" metric to fill in some boilerplate:

~~~scala
object MLMetric {
  val Invalid = MLMetric("Invalid", 0.0, false)
}
~~~

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -223,19 +223,13 @@ object MLParams {
...
- * @param trainingTime  (MLlib) Training time.
- *                      executionTime is set to the same value to match Spark Core tests.
- * @param trainingMetric  (MLlib) Training metric, such as accuracy
- * @param testTime  (MLlib) Test time (for prediction on test set, or on training set if there
- *                  is no test set).
- * @param testMetric  (MLlib) Test metric, such as accuracy
- * @param additionalTests  (MLlib) Additional methods test results.
+ * @param metricName the name of the metric
+ * @param metricValue the value of the metric
+ * @param isLargerBetter  (MLlib) Test metric, such as accuracy
```
mengxr: Doc doesn't match the param name.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -38,7 +40,7 @@ trait BenchmarkAlgorithm extends Logging {
   def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): Double = -1.0 // Not putting NaN because it is not valid JSON.
+      model: Transformer): MLMetrics = new MLMetrics()
```
mengxr: `MLMetric.Invalid` See comment below.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -67,9 +69,10 @@ trait ScoringWithEvaluator {
   final override def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): Double = {
+      model: Transformer): MLMetrics = {
     val eval = model.transform(testSet)
-    evaluator(ctx).evaluate(eval)
+    // TODO: add MetricName for evaluators
```
mengxr: Cannot leave this TODO here. We should check if evaluator hasMetricName, if yes, use the metric name, otherwise the class name of the evaluator.

src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
```diff
@@ -67,9 +69,10 @@ trait ScoringWithEvaluator {
...
+    new MLMetrics("", evaluator(ctx).evaluate(eval), evaluator(ctx).isLargerBetter)
```
mengxr: Code reads better if we move `evaluator` part out. It triggers an action.

~~~scala
val results = model.transform(testSet)
val eval = evaluator(ctx)
val metricName = ...
val metricValue = eval.evaluate(eval)
MLMetric(metricName, metricValue, eval.isLargerBetter)
~~~

* minor: case class doesn't need `new`

src/main/scala/com/databricks/spark/sql/perf/mllib/MLPipelineStageBenchmarkable.scala
```diff
@@ -72,9 +72,17 @@ class MLPipelineStageBenchmarkable(
       val (scoreTrainTime, scoreTraining) = measureTime {
         test.score(param, trainingData, model)
       }
+      val metricTrainTime = new MLMetrics("trainTime", scoreTrainTime.toMillis, false)
```
mengxr: Let's use "training" instead of "train" (verb). We can use `.` to separate `training`/`test` from the metric names. Say `training.accuracy`, `training.time`, `training.areaUnderROC`.  So we don't need to worry about keep camelCase ...

src/main/scala/com/databricks/spark/sql/perf/mllib/MLPipelineStageBenchmarkable.scala
```diff
@@ -86,12 +94,7 @@ class MLPipelineStageBenchmarkable(
           tuple._1 -> additionalMethodTime.toMillis.toDouble
       }
 
-      val ml = MLResult(
-        trainingTime = Some(trainingTime.toMillis),
-        trainingMetric = Some(scoreTraining),
-        testTime = Some(scoreTestTime.toMillis),
-        testMetric = Some(scoreTest / testDataCount.get),
-        additionalTests = additionalTests)
+      val ml = Array(metricTrainTime, metricTrain, metricTestTime, metricTest)
```
mengxr: `ml` -> `mlMetrics`

src/main/scala/com/databricks/spark/sql/perf/mllib/MLPipelineStageBenchmarkable.scala
```diff
@@ -72,9 +72,17 @@ class MLPipelineStageBenchmarkable(
       val (scoreTrainTime, scoreTraining) = measureTime {
         test.score(param, trainingData, model)
       }
+      val metricTrainingTime = MLMetric("training.time", scoreTrainTime.toMillis, false)
+      val metricTraining = MLMetric("training."+scoreTraining.metricName,
```
mengxr: minor: `val metricTraining = scoreTraining.copy(metricName = "training." + scoreMetric.name)`

PR #145, opened 2018-06-03
[ML-3753] Log "value" instead of "Some(value)" for ML params in results

ludatabricks:
Change the function MLParams.toMap so it will not output a option value in the map.
We will not get option value in the params in the output result.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/results.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/results.scala
@@ -83,7 +83,7 @@ case class BenchmarkResult(
     breakDown: Seq[BreakdownResult] = Nil,
     queryExecution: Option[String] = None,
     failure: Option[Failure] = None,
-    mlResult: Option[MLResult] = None)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -149,8 +149,8 @@ class MLParams(
    */
   def toMap: Map[String, String] = {
     val allParams = ReflectionUtils.getConstructorArgs(this)
-    allParams.map { case (key: String, value: Any) =>
-      key -> value.toString
+    allParams.map { case (key: String, value: Option[Any]) =>
+      key -> value.get.toString
```
mengxr: This will throw an exception if the value is `None` or `null`. Try this:

~~~scala
// Only outputs params that have values.
allParams.flatMap {
  case (key: String,   Some(value: Any)) =>
    Some(key -> value.toString)
  case _ =>
    None
}
~~~

PR #146, opened 2018-06-04
[ML-3775] Add "benchmarkId" to BenchmarkResult

ludatabricks:
benchmarkId is based on the benchmark name and a hashed value of params.


src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -83,7 +87,8 @@ case class BenchmarkResult(
     breakDown: Seq[BreakdownResult] = Nil,
     queryExecution: Option[String] = None,
     failure: Option[Failure] = None,
-    mlResult: Option[Array[MLMetric]] = None)
+    mlResult: Option[Array[MLMetric]] = None,
+    benchmarkId: String = null)
```
mengxr: `benchmarkId: Option[String] = None` to be consistent with other fields

src/main/scala/com/databricks/spark/sql/perf/results.scala
```diff
@@ -67,6 +67,10 @@ case class BenchmarkConfiguration(
  *               For example, it can be the number of rows generated by this query or
  *               the sum of hash values of rows generated by this query.
  * @param breakDown The breakdown results of the query plan tree.
+ * @param queryExecution The query execution plan.
+ * @param failure The failure message.
+ * @param mlResult The result metrics specific to MLlib.
+ * @param benchmarkId The unique benchmark ID.
```
mengxr: "Optionally an ID to identify a series of benchmark runs. In ML, this is generated based on the benchmark name and the hash value of params."

If you say "unique", you should state under what context it is unique. And we don't really guarantee uniqueness.

PR #147, opened 2018-06-07
[ML-3584] Added benchmarks to mllib-large.yaml for ALS

ludatabricks:
Benchmark for ALS is added to mllib-large.yaml.
Then parameters is based on the previous benchmarks for the Spark 2.2 QA. It has been tested for the stability under the same cluster setting for other benchmarks (classifcation).


mengxr:
LGTM. Merged.

PR #149, opened 2018-06-08
[ML-3585] Added benchmarks to mllib-large.yaml for clustering

ludatabricks:
Benchmark for clustering is added to mllib-large.yaml.
GaussianMixture, KMeans, and LDA are added. BisectingKMeans is missing in spark-sql-perf now. Need to be fixed in the following up JIRA: https://databricks.atlassian.net/browse/ML-3834
Then parameters is based on the previous benchmarks for the Spark 2.2 QA.


src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
```diff
@@ -37,6 +37,28 @@ benchmarks:
...
       smoothing: 1.0
+  - name: clustering.GaussianMixture
+    params:
+      numExamples: 100000
+      numTestExamples: 100000
+      numFeatures: 1000
+      k: 10
+      maxIter: 10
+      tol: 0.01
+  - name: clustering.KMeans
```
mengxr: what is the dataset size for kmeans?

ludatabricks: The size is 1000000.

mengxr: But why it is not a param?

ludatabricks: We have some common default parameters:
common:
  numExamples: 1000000
  numTestExamples: 1000000
  numFeatures: 4000
  numPartitions: 64
  randomSeed: [1, 1, 1]

These will be the params for the benchmark call, but we don't need to write it for each case.

mengxr:
LGTM

PR #150, opened 2018-06-08
[ML-3581] Add benchmarks to mllib-large.yaml for regression

ludatabricks:
Benchmark for regression is added to mllib-large.yaml.
DecisionTreeRegression, GLMRegression, LinearRegression, and RandomForestRegression are added. 

GBT, AFTSurvivalRegression, and IsotonicRegression are missing in spark-sql-perf.

### DIFF
--- a/src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
+++ b/src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
@@ -68,3 +68,26 @@ benchmarks:
       regParam: 0.01
       rank: 10
       maxIter: 10
+  - name: regression.DecisionTreeRegression
+    params:
+      depth: [5, 10]
+      numClasses: 4
+  - name: regression.GLMRegression
+    params:
+      numExamples: 500000
+      numTestExamples: 500000
+      numFeatures: 1000
+      link: log
+      family: gaussian
+      tol: 1e-6
+      maxIter: 10
+      regParam: 0.1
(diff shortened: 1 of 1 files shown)


src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
```diff
@@ -67,4 +67,27 @@ benchmarks:
       numItems: 6000000
       regParam: 0.01
       rank: 10
-      maxIter: 10
\ No newline at end of file
+      maxIter: 10
+  - name: regression.DecisionTreeRegression
+    params:
+      depth: [5, 10]
+      numClasses: 4
```
jkbradley: Regression should not use numClasses

src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
```diff
@@ -67,4 +67,27 @@ benchmarks:
...
+  - name: regression.GLMRegression
+    params:
+      numExamples: 500000
+      numTestExamples: 500000
+      numFeatures: 1000
+      link: log
+      family: gaussian
+      tol: 1e-6
```
jkbradley: If we're testing for time/iteration, then I'd set tol to 0 and use maxIter to set the iteration count.

jkbradley:
The running times look fine.  Thanks!

jkbradley:
LGTM
Merging with master
Thanks!

PR #139, opened 2018-04-04
Additional method test for some ML algos

WeichenXu123:
Add additional method test for some ML algos.

In this PR, I add `associationRules` in `FPGrowth` and `findSynonyms`. 
After the design is accepted, I will add other methods later.

Add an interface in `BenchmarkableAlgorithm`:
```
  def testAdditionalMethods(ctx: MLBenchContext, model: Transformer): Map[String, () => _]
```

PR #151, opened 2018-06-09
[ML-3824] Add benchmarks to mllib-large.yaml for FPGrowth

ludatabricks:
Benchmark for FPGrowth is added to mllib-large.yaml.

### DIFF
--- a/src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
+++ b/src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
@@ -59,6 +59,10 @@ benchmarks:
       optimizer:
         - em
         - online
+  - name: fpm.FPGrowth
+    params:
+      numItems: 10000
+      itemSetSize: 10
   - name: recommendation.ALS
     params:
       numExamples: 50000000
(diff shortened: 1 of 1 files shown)


src/main/resources/com/databricks/spark/sql/perf/mllib/config/mllib-large.yaml
```diff
@@ -59,6 +59,10 @@ benchmarks:
       optimizer:
         - em
         - online
+  - name: fpm.FPGrowth
+    params:
+      numItems: 10000
+      itemSetSize: 10
```
jkbradley: This is a pretty big average itemset size.  It could be worth testing 2 sizes, say 4 and 10.  (Think of this number as being, e.g., the number of items in an online shopping cart at checkout.)

jkbradley:
@ludatabricks  and @mengxr I see you removed logging of metrics for additional methods like AssociationRules in this PR: https://github.com/databricks/spark-sql-perf/pull/144/files

This logging was added in https://github.com/databricks/spark-sql-perf/pull/139/files so that we could time things other than fit() and transform().

Was this a mistake, or was there a problem with that logging?

ludatabricks:
@jkbradley. That is a mistake. I should leave a comment there. In the new schema, we only want to save the results of the new methods instead of the function. I will fix it in a separate PR.

jkbradley:
Sure, I'll say this LGTM, and we can adjust running times later as needed.  Thanks!

PR #153, opened 2018-06-12
[ML-3915] add additionalTests to MLMetrics

ludatabricks:
The time for additionalTests is missing in MLMetrics. Now add it back to MLMetrics so that we can test the time for other methods.


jkbradley:
LGTM
Thanks!
Merging with master

PR #155, opened 2018-06-14
[ML-3870] Make spark-sql-perf master compiled with spark 2.3 and scala 2.11

ludatabricks:
Change the build config to update spark 2.3 and update the scala dependence in bin/spark-perf


mengxr:
LGTM2. Merged.

PR #158, opened 2018-07-03
Update VectorAssembler test such that the dataset size is numExamples * numFeatures

mengxr:
Currently the dataset size is numExamples * numFeatures * numInputCols, which is much bigger than other ML perf tests. This PR updates its implementation and makes it more efficient at slicing vectors.

Tested on the mllib-big.yaml and 3 runs finished in < 2 minutes.

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/feature/VectorAssembler.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/feature/VectorAssembler.scala
@@ -13,35 +13,41 @@ import com.databricks.spark.sql.perf.mllib.{BenchmarkAlgorithm, MLBenchContext, 
 /** Object for testing VectorAssembler performance */
 object VectorAssembler extends BenchmarkAlgorithm with TestFromTraining {
 
+  private def getInputCols(numInputCols: Int): Array[String] = {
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/spark/sql/perf/mllib/feature/VectorAssembler.scala
```diff
@@ -13,35 +13,41 @@ import com.databricks.spark.sql.perf.mllib.{BenchmarkAlgorithm, MLBenchContext,
...
-      numFeatures * numInputCols
-    )
-    val sliceVec = udf { (v: Vector, from: Int, until: Int) =>
-      Vectors.dense(v.toArray.slice(from, until))
-    }
-    for (i <- (1 to numInputCols.get)) {
-      val colName = s"inputCol${i.toString}"
-      val fromIndex = (i - 1) * numFeatures
-      val untilIndex = i * numFeatures
-      df = df.withColumn(colName, sliceVec(col("features"), lit(fromIndex), lit(untilIndex)))
```
mengxr: The previous implementation slices the vector multiple times, which is slow. The new implementation slices the vector only once.

jkbradley:
LGTM
Merging with master

PR #159, opened 2018-07-03
[ML-2918] Call count() in default score() to improve timing of transform()

jkbradley:
For Models and Transformers which are not tested with Evaluators, I think we are not timing transform() correctly here: https://github.com/databricks/spark-sql-perf/blob/aa1587fec5a08eeca78a06389599939d8dced0bf/src/main/scala/com/databricks/spark/sql/perf/mllib/MLPipelineStageBenchmarkable.scala#L65

Since transform() is lazy, we need to materialize it during timing.  This PR currently just calls count() in the default implementation of score().

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
@@ -35,12 +35,17 @@ trait BenchmarkAlgorithm extends Logging {
   /**
    * The unnormalized score of the training procedure on a dataset. The normalization is
    * performed by the caller.
+   * This calls `count()` on the transformed data to attempt to materialize the result for
+   * recording timing metrics.
    */
   @throws[Exception]("if scoring fails")
   def score(
       ctx: MLBenchContext,
       testSet: DataFrame,
-      model: Transformer): MLMetric = MLMetric.Invalid
+      model: Transformer): MLMetric = {
+    model.transform(testSet).count()
+    MLMetric.Invalid
+  }
 
   def name: String = {
     this.getClass.getCanonicalName.replace("$", "")
--- a/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/mllib/BenchmarkAlgorithm.scala
@@ -43,7 +43,11 @@ trait BenchmarkAlgorithm extends Logging {
       ctx: MLBenchContext,
       testSet: DataFrame,
       model: Transformer): MLMetric = {
-    model.transform(testSet).count()
+    val output = model.transform(testSet)
+    // We create a useless UDF to make sure the entire DataFrame is instantiated.
+    val fakeUDF = udf { (_: Any) => 0 }
+    val columns = testSet.columns
+    output.select(sum(fakeUDF(struct(columns.map(col) : _*)))).first()
     MLMetric.Invalid
   }
 


jkbradley:
@mengxr  Do you think that count() might be optimized too much by Catalyst and skip the transform?  If so, we could call cache().count(), though that might overestimate the cost of transform().

mengxr:
I don't think Spark SQL knows about the statistics in this case because the datasets are converted from RDD, not from `range()` or files on disk.

jkbradley:
I like the UDF idea, but tailoring this for different output column names seems like overkill to me for now.  I'll test the current PR on mllib-large and report back.

mengxr:
It can be nesting all columns under a single column and send it to the dummy UDF, if count() does not always work.

jkbradley:
OK, I guess that works, though it might be an overestimate like caching.

I did confirm that the current PR changes test times from 0 to non-0 values, so it does something.  : P

But I can update it to do the UDF thing.

mengxr:
Just checked Dataset.count(), which uses Dataset().groupBy().count().collect().head.getLong(0). And the query plan of df.groupBy().count() is:

~~~
spark.range(100000).groupBy().count().explain()

*(2) HashAggregate(keys=[], functions=[finalmerge_count(merge count#132783L) AS count(1)#132777L])
+- Exchange SinglePartition
   +- *(1) HashAggregate(keys=[], functions=[partial_count(1) AS count#132783L])
      +- *(1) Project
         +- *(1) Range (0, 100000, step=1, splits=292)
~~~

So it doesn't materialize any columns. Let's go with the UDF approach.

jkbradley:
Yeah, I think you're right.  I checked with the actual code, and it uses the data generated before the prediction to do the count:
```
== Parsed Logical Plan ==
Aggregate [count(1) AS count#1055L]
+- AnalysisBarrier
      +- Project [items#962, UDF(items#962) AS prediction#1049]
         +- Project [_1#960 AS items#962]
            +- SerializeFromObject [mapobjects(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), staticinvoke(class org.apache.spark.unsafe.types.UTF8String, StringType, fromString, lambdavariable(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), true), true, false), assertnotnull(assertnotnull(input[0, scala.Tuple1, true]))._1, None) AS _1#960]
               +- ExternalRDD [obj#959]

== Analyzed Logical Plan ==
count: bigint
Aggregate [count(1) AS count#1055L]
+- Project [items#962, UDF(items#962) AS prediction#1049]
   +- Project [_1#960 AS items#962]
      +- SerializeFromObject [mapobjects(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), staticinvoke(class org.apache.spark.unsafe.types.UTF8String, StringType, fromString, lambdavariable(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), true), true, false), assertnotnull(assertnotnull(input[0, scala.Tuple1, true]))._1, None) AS _1#960]
         +- ExternalRDD [obj#959]

== Optimized Logical Plan ==
Aggregate [count(1) AS count#1055L]
+- Project
   +- InMemoryRelation [items#962], true, 10000, StorageLevel(disk, memory, deserialized, 1 replicas)
         +- *(1) Project [_1#960 AS items#962]
            +- *(1) SerializeFromObject [mapobjects(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), staticinvoke(class org.apache.spark.unsafe.types.UTF8String, StringType, fromString, lambdavariable(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), true), true, false), assertnotnull(input[0, scala.Tuple1, true])._1, None) AS _1#960]
               +- Scan ExternalRDDScan[obj#959]

== Physical Plan ==
*(2) HashAggregate(keys=[], functions=[finalmerge_count(merge count#1063L) AS count(1)#1054L], output=[count#1055L])
+- Exchange SinglePartition
   +- *(1) HashAggregate(keys=[], functions=[partial_count(1) AS count#1063L], output=[count#1063L])
      +- InMemoryTableScan
            +- InMemoryRelation [items#962], true, 10000, StorageLevel(disk, memory, deserialized, 1 replicas)
                  +- *(1) Project [_1#960 AS items#962]
                     +- *(1) SerializeFromObject [mapobjects(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), staticinvoke(class org.apache.spark.unsafe.types.UTF8String, StringType, fromString, lambdavariable(MapObjects_loopValue22, MapObjects_loopIsNull22, ObjectType(class java.lang.String), true), true, false), assertnotnull(input[0, scala.Tuple1, true])._1, None) AS _1#960]
                        +- Scan ExternalRDDScan[obj#959]
```

jkbradley:
I verified that the UDF works.  It takes longer, and the plan shows the UDF being used: https://dogfood.staging.cloud.databricks.com/sparkui/0706-175109-matt861/driver-2449086398155064347/SQL/execution?id=36&o=0
https://dogfood.staging.cloud.databricks.com/sparkui/0706-175109-matt861/driver-2449086398155064347/SQL/execution?id=35&o=0

mengxr:
LGTM

PR #162, opened 2018-07-09
[ML-4154] Added testing for before/after of ml benchmarks.

MrBago:
This PR adds a unit tests which runs the `beforeBenchmark` & `afterBenchmark` methods for the benchmarks included in mllib-small.yaml.

### DIFF
--- a/src/test/scala/com/databricks/spark/sql/perf/mllib/MLLibSuite.scala
+++ b/src/test/scala/com/databricks/spark/sql/perf/mllib/MLLibSuite.scala
@@ -2,27 +2,28 @@ package com.databricks.spark.sql.perf.mllib
 
 import org.scalatest.{BeforeAndAfterAll, FunSuite}
 
-import org.apache.spark.{SparkConf, SparkContext}
-import org.apache.spark.sql.Row
+import org.apache.spark.sql.{Row, SparkSession}
 
 class MLLibSuite extends FunSuite with BeforeAndAfterAll {
 
-  private var sc: SparkContext = _
+  private var sparkSession: SparkSession = _
 
   override def beforeAll(): Unit = {
     super.beforeAll()
-    val sparkConf = new SparkConf().setAppName("MLlib QA").setMaster("local[2]")
-    sc = SparkContext.getOrCreate(sparkConf)
+    sparkSession = SparkSession.builder.master("local[2]").appName("MLlib QA").getOrCreate()
   }
 
   override def afterAll(): Unit = {
-    super.afterAll()
-    if (sc != null) {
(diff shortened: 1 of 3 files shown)


src/test/scala/com/databricks/spark/sql/perf/mllib/MLLibSuite.scala
```diff
@@ -1,10 +1,29 @@
...
+  private var sc: SparkContext = _
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    val sparkConf = new SparkConf().setAppName("MLlib QA").setMaster("local[2]")
+    sc = SparkContext.getOrCreate(sparkConf)
+  }
+
+  override def afterAll(): Unit = {
+    super.afterAll()
```
jkbradley: Call super.afterAll at the end of this method

src/test/scala/com/databricks/spark/sql/perf/mllib/MLLibSuite.scala
```diff
@@ -1,10 +1,29 @@
...
 
-class MLLibTest extends FunSuite {
+class MLLibSuite extends FunSuite with BeforeAndAfterAll {
+
+  private var sc: SparkContext = _
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    val sparkConf = new SparkConf().setAppName("MLlib QA").setMaster("local[2]")
+    sc = SparkContext.getOrCreate(sparkConf)
```
jkbradley: This is fine with me since it works, but you could use a SparkSession since we don't need to support Spark 2.0 anymore.

src/test/scala/com/databricks/spark/sql/perf/mllib/MLLibSuite.scala
```diff
@@ -1,10 +1,29 @@
...
+  }
+
+  override def afterAll(): Unit = {
+    super.afterAll()
+    if (sc != null) {
+      sc.stop()
+    }
+    // To avoid RPC rebinding to the same port, since it doesn't unbind immediately on shutdown
```
jkbradley: curious: Is this from somewhere or because of a problem you hit?

MrBago: I copied this from the `LocalSparkSession` trait, https://livegrep.dev.databricks.com/view/databricks/spark-master/core/src/test/scala/org/apache/spark/LocalSparkContext.scala#L55

jkbradley:
LGTM
Merging
Thanks!

PR #164, opened 2018-08-17
Fix compile for Spark 2.4 SNAPSHOT and only catch NonFatal

mengxr:
The build is currently broken with Spark 2.4. This PR fixes the compile. It won't pass CI since 2.4.0 SNAPSHOT is not available on maven. I will leave the PR here until RC is out.

* only catch NonFatal instead of Throwable. Our job hanged on Spark 2.4 cluster due to class compatibility issue. I'm not sure if catching Throwable is the root cause here. Let me know if we do want to catch `Throwable`.
* Remove `afterBenchmark` from MLlib. The API no longer exists in Spark.
* Update tree generation code, because it uses some internal APIs that we changed in 2.4.

cc: @MrBago @juliuszsompolski

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -44,6 +44,8 @@ libraryDependencies += "org.yaml" % "snakeyaml" % "1.17"
 
 libraryDependencies += "com.typesafe.scala-logging" %% "scala-logging-slf4j" % "2.1.2"
 
+resolvers += "Apache Development Snapshot Repository" at "https://repository.apache.org/content/repositories/snapshots"
+
 fork := true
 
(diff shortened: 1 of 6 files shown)


mengxr:
Run succeeded with the new jar: https://dogfood.staging.cloud.databricks.com/#job/830162/run/80.

juliuszsompolski:
cc @npoggi

mengxr:
@juliuszsompolski @npoggi I added Apache snapshot repo to resolvers. Shall we merge this patch? It will help 2.4 QA.

PR #165, opened 2018-08-19
Fix 3 local benchmark classes

g1thubhub:
This PR fixes broken benchmarks so that 'AggregationPerformance' and 'JoinPerformance' can be run (they don't work currently) and the average benchmark for Datasets in 'DatasetPerformance' work. I ran all benchmarks with those fixes successfully and all changes below are necessary -- for example, without the explicit type annotation ": Seq[Benchmarkable]" the benchmarks did not work on my computer likely due to some compiler/reflection hiccups

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ The first run of `bin/run` will build the library.
 
 Use `sbt package` or `sbt assembly` to build the library jar.
 
+## Local performance tests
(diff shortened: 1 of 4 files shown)


juliuszsompolski:
cc @npoggi

g1thubhub:
Added an explanation and usage info to the ReadMe, let me know if something is still missing or should be changed

npoggi:
Merging, Travis test errors are related to 2.4-SNAPSHOT availability

PR #166, opened 2018-09-10
Make TPCDS queryNames public so it can be accessed from notebooks.

npoggi:
Just a small change so that `queryNames` val from `com/databricks/spark/sql/perf/tpcds/TPCDS_2_4_Queries.scala` is accessible from notebooks and get the internal query naming to select individual queries.


winglungngai:
LGTM

PR #170, opened 2018-09-26
wait for all broadcast being cleaned after each benchmark

cloud-fan:
The broadcasted data stay in the executor block manager and make executor have less execution memory. We should wait it to be cleaner before running next benchmark


cloud-fan:
@npoggi do we have jenkins for this project?

npoggi:
@cloud-fan we use travis-ci for this project.  Why dit you closed this PR?

cloud-fan:
For the perf regression, the problem is task skew not broadcast, so this patch is not useful now.

PR #174, opened 2018-10-31
[ML-5437] Build with spark-2.4-rc and resolve build issues.

MrBago:
We made some changes to related to new APIs in spark2.4. These APIs were reverted because they were breaking changes so we need to revert our changes.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -14,7 +14,7 @@ sparkPackageName := "databricks/spark-sql-perf"
 // All Spark Packages need a license
 licenses := Seq("Apache-2.0" -> url("http://opensource.org/licenses/Apache-2.0"))
 
-sparkVersion := "2.4.0-SNAPSHOT"
+sparkVersion := "2.4.0"
 
 sparkComponents ++= Seq("sql", "hive", "mllib")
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -44,7 +44,9 @@ libraryDependencies += "org.yaml" % "snakeyaml" % "1.17"
 
 libraryDependencies += "com.typesafe.scala-logging" %% "scala-logging-slf4j" % "2.1.2"
 
-resolvers += "Apache Development Snapshot Repository" at "https://repository.apache.org/content/repositories/snapshots"
+// Use RC5 until public release is available
```
npoggi: Should we just wait a couple more days for 2.4 to be out?

MrBago: I just updated it to use the official release.

src/test/scala/com/databricks/spark/sql/perf/mllib/MLLibSuite.scala
```diff
@@ -2,18 +2,31 @@ package com.databricks.spark.sql.perf.mllib
...
 class MLLibSuite extends FunSuite with BeforeAndAfterAll {
 
   private var sparkSession: SparkSession = _
+  var savedLevels: Map[String, Level] = _
 
   override def beforeAll(): Unit = {
     super.beforeAll()
     sparkSession = SparkSession.builder.master("local[2]").appName("MLlib QA").getOrCreate()
+
+    savedLevels = Seq("akka", "org", "com.databricks").map { name =>
```
jkbradley: What's the deal with this?  Was this test logging too much and breaking Travis?  If this is needed, how about adding a note here saying what the issue was?

MrBago: Ya, the logs were causing travis to exit with, "The job exceeded the maximum log length, and has been terminated." I'll add a comment.

PR #180, opened 2019-05-27
Fix files truncating according to maxRecordPerFile

gczsjdy:
`(numRows)/maxRecordPerFile)` automatically performs a `floor`, and invalidate the `ceil` : )

### DIFF
--- a/src/main/scala/com/databricks/spark/sql/perf/Tables.scala
+++ b/src/main/scala/com/databricks/spark/sql/perf/Tables.scala
@@ -222,7 +222,7 @@ abstract class Tables(sqlContext: SQLContext, scaleFactor: String,
           log.info(s"Data has $numRows rows clustered $clusterByPartitionColumns for $maxRecordPerFile")
 
           if (maxRecordPerFile > 0 && numRows > maxRecordPerFile) {
-            val numFiles = ((numRows * 1.0)/maxRecordPerFile).ceil.toInt
+            val numFiles = (numRows.toDouble/maxRecordPerFile).ceil.toInt
             println(s"Coalescing into $numFiles files")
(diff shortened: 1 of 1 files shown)


gczsjdy:
cc @yhuai @cloud-fan It's a very simple change, could you please take a look?

src/main/scala/com/databricks/spark/sql/perf/Tables.scala
```diff
@@ -222,7 +222,7 @@ abstract class Tables(sqlContext: SQLContext, scaleFactor: String,
           log.info(s"Data has $numRows rows clustered $clusterByPartitionColumns for $maxRecordPerFile")
 
           if (maxRecordPerFile > 0 && numRows > maxRecordPerFile) {
-            val numFiles = ((numRows)/maxRecordPerFile).ceil.toInt
+            val numFiles = ((numRows * 1.0)/maxRecordPerFile).ceil.toInt
```
cloud-fan: is it better to call `numRows.toDouble`?

gczsjdy: Done.

cloud-fan:
thanks merged!

npoggi:
@gczsjdy thanks for spotting it.  That code could trigger only at 10TB+ scale.

gczsjdy:
@npoggi No problem. : )
Why? I think it can also trigger below 10T scale, although we are benchmarking on a 10T scale.

npoggi:
@gczsjdy due to the default `maxRecordPerFile`, with the current partition scheme, it should not be reached at smaller scales.  Anyway, other other improvements please submit a PR :)

PR #187, opened 2019-11-28
Spark3 tpcds setup

Peach-He:
Noticed that spark 3.0.0 does not have method org.apache.spark.sql.SQLContext.createExternalTable, updated the TPC-DS setup part to support spark 3.0.0 by using SparkSession to create table.
Thanks for your reviewing.


isamllr:
Tested... unfortunately, this won't build if I change the Spark version to 3.0.0 in built.sbt.

PR #201, opened 2021-04-26
Update the TPCDS schema based on the Spark codebase

maropu:
This PR intends to update the TPCDS schema based on [the Spark codebase](https://github.com/apache/spark/blob/master/sql/core/src/test/scala/org/apache/spark/sql/TPCDSBase.scala).

I've manually checked that tables can be generated in spark-3.0/3.1/master.


src/main/scala/com/databricks/spark/sql/perf/tpcds/TPCDSTables.scala
```diff
@@ -59,484 +56,554 @@ class TPCDSTables(
...
-  extends Tables(sqlContext, scaleFactor, useDoubleForDecimal, useStringForDate) {
-  import sqlContext.implicits._
+  useStringForDate: Boolean = false,
+  useStringForCharVarchar: Boolean = true)
+  extends Tables(sqlContext, scaleFactor, useDoubleForDecimal, useStringForDate, useStringForCharVarchar) {
 
   val dataGenerator = new DSDGEN(dsdgenDir)
+
+ /**
+  * Datatype mapping for TPC-DS and Spark SQL, fully matching schemas defined in `tpcds.sql` of the
```
maropu, line 66: This comment was copied from `TPCDSBase` in Spark. cc: @yaooqinn

src/main/scala/com/databricks/spark/sql/perf/tpcds/TPCDSTables.scala
```diff
@@ -59,484 +56,554 @@ class TPCDSTables(
...
-      'cs_ext_tax               .decimal(7,2),
-      'cs_coupon_amt            .decimal(7,2),
-      'cs_ext_ship_cost         .decimal(7,2),
-      'cs_net_paid              .decimal(7,2),
-      'cs_net_paid_inc_tax      .decimal(7,2),
-      'cs_net_paid_inc_ship     .decimal(7,2),
-      'cs_net_paid_inc_ship_tax .decimal(7,2),
-      'cs_net_profit            .decimal(7,2)),
+      """
+        |`cs_sold_date_sk` INT,
```
maropu, line 91: Since `ColumnName` does not have a helper method for char/varchar types, I've replaced schema fields with a string schema.

maropu:
cc: @HyukjinKwon @npoggi

npoggi:
Thanks for the update, I'll take a closer look later during the week.

PR #218, opened 2025-04-14
add tpcds-v4.0 support

heyujiao99:
- Add support for TPCDS-v4.0.0
    - add queries generated by the official TPC-DS-Tools-v4.0.0 ( [tpc.org](https://www.tpc.org/tpc_documents_current_versions/current_specifications5.asp))
    - add entries for tpcds-v4.0.0 in TPCDS_4_0_Queries.scala and TPCDS.scala
    - since v2.6.0 the field 'c_last_review_date' was renamed to 'c_last_review_date_sk' (used in q30). For backward compatibility, changed it back to 'c_last_review_date' in q30
- The tpcds-kit compatible with v4.0.0 can be found at [tpcds-kit](https://github.com/heyujiao99/tpcds-kit/tree/v4.0.0)
- Tested on Linux platform with 1TB dataset.


mwc360:
All of the old spec docs I've looked at reference 'c_last_review_date**_sk**' and have this as an identity column which would map to a int, not a string. I haven't found any other instances of this column being used this way, I've only seen it in this repo. Might it be worth just updating this tool to generate the proper int field w/ '_sk'?

